PR #10, opened 2023-07-30
[FEATURE]: Added debian and archive releases

var77:
## Description
- Add packaging with `cpack` and upload the packaged `.deb` file to action output artifacts.
- Added archive packaging which will create `tar` archive.

The action output will contain `deb` packages and `tar` universal package as shown in screenshot
Currently only `amd64` packages are being generated

## Installation

To install from `deb` package, download the package and run

```
dpkg -i lanterndb-0.0.2-postgres-15-amd64.deb
```
Assuming you have already installed `postgres` and `pgvector` of the specified version

To install from universal archive package download the tar archive and run

```
tar xvf lanterndb-0.0.2.tar && \
cd lanterndb-0.0.2 && \ 
make install
```

[Example action run](https://github.com/lanterndata/lanterndb/actions/runs/5705993263)

image


.github/workflows/build-and-test-linux.yaml
```diff
@@ -30,14 +30,30 @@ jobs:
...
+        id: build
+        run: sudo su -c "PG_VERSION=$PG_VERSION USE_SOURCE=1 GITHUB_OUTPUT=$GITHUB_OUTPUT ./ci/scripts/build-linux.sh"
         env:
           PG_VERSION: ${{ matrix.postgres }}
           BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+      - name: Create Archive Package
+        id: archive
+        run: sudo su -c "GITHUB_OUTPUT=$GITHUB_OUTPUT ./ci/scripts/package-archive.sh"
+        if: ${{ matrix.postgres == '15' }} # The output is not dependent to pg version, so we can run this only once
       # Enable tmate debugging of manually-triggered workflows if the input option was provided
```
Ngalstyan4, line 42: Is this to be added later?
add a "todo::", if so.

.github/workflows/build-and-test-linux.yaml
```diff
@@ -30,14 +30,30 @@ jobs:
...
       # - name: Setup tmate session
       #   uses: mxschmitt/action-tmate@v3
       - name: Run tests
         run: sudo su postgres -c "PG_VERSION=$PG_VERSION ./ci/scripts/run-tests.sh"
         env:
           PG_VERSION: ${{ matrix.postgres }}
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
```
Ngalstyan4, line 50: Will these run when someone forks the repo? doe these use any secrets?
We should properly gate any workflows that use secrets so people do not get annoying build fail emails after forking our repository

CMakeLists.txt
```diff
@@ -180,3 +180,19 @@ add_custom_target(
   COMMAND ${CMAKE_SOURCE_DIR}/scripts/bench.sh
   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
  )
+
+# Package universal install script
+add_custom_target(
+  archive
+  ${CMAKE_COMMAND} -E env SOURCE_DIR=${CMAKE_SOURCE_DIR} BUILD_DIR=${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/scripts/package.sh
+  DEPENDS ${CMAKE_BINARY_DIR}/lanterndb.so
```
Ngalstyan4, line 188: This depends on the whole target (the control file, the generated SQL files) and not just the shared library, right?
I think you can do `add_dependency(archive lanterndb)` to better capture this.
I am just afraid that if we do something that only modifies an SQL file, and already had an old build, `make archive` will not re-trigger a build.

cmake/PackageExtensionTemplate.cmake
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 3.3)
+project(LanternDB)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
+find_package(PostgreSQL REQUIRED)
+
+install(DIRECTORY ${PROJECT_SOURCE_DIR}/src/
```
Ngalstyan4, line 7: Is this a standalone cmake project file that will be invoked separately?
Why are these not in the maim CMakeLists.txt?

I think it is typical that `cmake/` folder only contains helper modules/function for the main CMakeLists file. But maybe I am missing something here.

Ngalstyan4 (commented):
Added some questions/notes

PR #18, opened 2023-08-03
[FEATURE]: support real[] array type for hnsw index

var77:
## Description
With the PR, we are making `pgvector` dependency optional by adding support for `real[]` type.
- Added `Dockerfile.dev` which will install the extension in debug build and also install `gdb` which will help us on debugging.
- Updated `sql` files to enable `l2_distance` function for `real[]` type, create `<->` operator for `real[]` type and also conditionally enable operators for `vector` type only if the type exsits.
- Added options `dims` on `hnsw` index, which is only required if index is being built on `real[]` type. This will indicate the dimension of the vectors..
- Added `l2_distance` function in `hnsw.c` which will calculate l2 distance between 2 arrays.
- Added `GetIndexDataType` function which will determine the data type in which the index was built from its oid and return value of `HnswDataType` enum (`REAL_ARRAY`, `VECTOR`, `UNKNOWN`). This will help us in cases where we need to change behavior based on index data type.
- Added `GetHnswIndexDimensions` function which will check if indexed data type is `REAL_ARRAY` return the dimension from options, if it is vector it will return from `pgvector` exposed metadata.
- Added `CheckHnswIndexDimensions` function, which will work only for `REAL_ARRAY` types and check if the length of provided array is equal to the dimension specified in index. It will throw error in case of mismatch.
- Added `hnsw_insert_array.sql` test case which is copy of `hnsw_insert.sql` test case (with some exclusions). This will build the index on type `real[]` instead of vector, and  test against it.

## TODO

- [x] Add more test cases to cover type chekings


src/hnsw/build.c
```diff
@@ -104,20 +173,21 @@ static void InitBuildState(HnswBuildState *buildstate, Relation heap, Relation i
...
-    buildstate->dimensions = TupleDescAttr(index->rd_att, 0)->atttypmod;
+    HnswDataType columnType = GetIndexDataType(index);
 
-    /* Require column to have dimensions to be indexed */
-    if(buildstate->dimensions < 0) elog(ERROR, "column does not have dimensions");
+    buildstate->dimensions = GetHnswIndexDimensions(index);
 
-    // todo:: check here that type of column is vector
+    /* Require column to have dimensions to be indexed */
+    if(buildstate->dimensions < HNSW_MIN_DIMS) elog(ERROR, "column does not have dimensions");
```
Ngalstyan4, line 181: Hardcoding 1 is fine here. the error message gives away it is one anyway

Ngalstyan4 (commented):
Great work!
some nits, otherwise looks good.

src/hnsw/distfunc.c
```diff
@@ -0,0 +1,15 @@
+#include "postgres.h"
+
+#include "distfunc.h"
+
+#include "math.h"
+
+extern dist_t l2_dist_impl(coord_t const* ax, coord_t const* bx, size_t dim)
```
var77, line 7: change to l2_squared and remove `sqrtf(distance);` change name to `l2_sq_dist`

sql/lanterndb.sql
```diff
@@ -7,9 +7,46 @@ CREATE ACCESS METHOD hnsw TYPE INDEX HANDLER hnsw_handler;
...
--- taken from pgvector so our index can work with pgvector types
-CREATE OPERATOR CLASS vector_l2_ops
-	DEFAULT FOR TYPE vector USING hnsw AS
-	OPERATOR 1 <-> (vector, vector) FOR ORDER BY float_ops,
-	FUNCTION 1 vector_l2_squared_distance(vector, vector),
-	FUNCTION 3 l2_distance(vector, vector);
+
+-- functions
+
+CREATE FUNCTION l2_distance(real[], real[]) RETURNS real
```
var77, line 13: change name to `l2_sq_dist`

sql/lanterndb.sql
```diff
@@ -7,9 +7,46 @@ CREATE ACCESS METHOD hnsw TYPE INDEX HANDLER hnsw_handler;
...
+	SELECT EXISTS (
+    	SELECT 1
+    	FROM pg_type
+    	WHERE typname = 'vector'
+	) INTO type_exists;
+
+	IF type_exists THEN
+	-- The type exists
+	-- taken from pgvector so our index can work with pgvector types
+		CREATE OPERATOR CLASS vector_l2_ops
```
var77, line 44: add `if not exists`

PR #25, opened 2023-08-05
create index with dynamically allocated blockmaps

davkhech:
In this PR changing the logic how we allocate blockmap pages
first we allocate one blockmap page and then fill in data pages, when there's not enough space we allocate two more blockmap pages then again data pages, and so on
the number of blockmap pages increase by power of two

here's a little schema how blocks look like

hdr page -> blockmap page 1 -> data page 1, data page 2 ... data page gk -> blockmap page 2, blockmap page 3 -> data page gk + 1, data page gk + 2, .... data page gk + gf -> blockmap page 4 ... blockmap page 7 -> data page .... and so on

### DIFF
--- a/src/hnsw/external_index.c
+++ b/src/hnsw/external_index.c
@@ -15,8 +15,11 @@
 #include "usearch.h"
 #include "utils/relcache.h"
 
-static inline void fill_blockno_mapping(BlockNumber *wal_retriever_block_numbers);
-static Cache       wal_retriever_block_numbers_cache;
+static Cache  wal_retriever_block_numbers_cache;
+static uint32 g_blockmap_groups = 0;
+static uint32 g_blockmap_group_indexes[ HNSW_MAX_BLOCKMAP_GROUPS ];
+static uint32 g_number_of_blocks = 1;
+static uint32 g_num_vectors = 0;
 
 int UsearchNodeBytes(usearch_metadata_t *metadata, int vector_bytes, int level)
 {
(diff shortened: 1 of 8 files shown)


src/hnsw/external_index.c
```diff
@@ -16,6 +16,9 @@
 
 static inline void  fill_blockno_mapping();
 static BlockNumber *wal_retriever_block_numbers = NULL;
+static uint32 g_blockmap_groups = 0;
+static uint32 g_blockmap_group_indexs[ 32 ];
```
Ngalstyan4, line 20: indexs -> indexes 
`g_blockmap_group_indexes`

src/hnsw/external_index.c
```diff
@@ -46,37 +49,69 @@ static char *extract_node(char               *data,
...
+        Page page = GenericXLogRegisterBuffer(state, buf, GENERIC_XLOG_FULL_IMAGE);
+        PageInit(page, BufferGetPageSize(buf), sizeof(HnswIndexPageSpecialBlock));
+
+        HnswIndexPageSpecialBlock *special = (HnswIndexPageSpecialBlock *)PageGetSpecialPointer(page);
+        special->firstId = first_node_index + blockmap_id * HNSW_BLOCKMAP_BLOCKS_PER_PAGE;
+        special->nextblockno = InvalidBlockNumber;
+
+        HnswBlockmapPage *blockmap = palloc0(BLCKSZ);
+
+        memset(blockmap, 0, BLCKSZ);
```
Ngalstyan4, line 80: `palloc` just allocates memory. `palloc0` allocates it and sets it to zero
So no need for another memset here.

src/hnsw/external_index.c
```diff
@@ -46,37 +49,69 @@ static char *extract_node(char               *data,
...
+        if(PageAddItem(page, (Item)blockmap, sizeof(HnswBlockmapPage), InvalidOffsetNumber, false, false)
+           == InvalidOffsetNumber) {
+            // we always add a single Blockmap page per Index page which has a fixed size that
+            // always fits in postgres wal page. So this should never happen
+            // (Assumes 8k BLKSZ. we can make HnswBlockmapPage size configurable by BLCKSZ)
+            elog(ERROR, "could not add blockmap to page");
+        }
+
+        special->lastId = first_node_index + (blockmap_id + 1) * HNSW_BLOCKMAP_BLOCKS_PER_PAGE - 1;
+        special->nextblockno = BufferGetBlockNumber(buf) + 1;
```
Ngalstyan4, line 92: Is this valid for the last iteration of the for loop?

src/hnsw/external_index.c
```diff
@@ -46,37 +49,69 @@ static char *extract_node(char               *data,
...
+                        size_t          num_added_vectors) 
 {
+    // Create empty blockmap pages for the group
+    const int number_of_blockmaps_in_group = 1 << g_blockmap_groups;
+    elog(INFO, "davkhech number_of_blockmaps_in_group %d num_added_vectors %d", number_of_blockmaps_in_group, num_added_vectors);
+    for (int blockmap_id = 0; blockmap_id < number_of_blockmaps_in_group; ++blockmap_id) {
+        Buffer buf = ReadBufferExtended(index, forkNum, P_NEW, RBM_NORMAL, NULL);
+        LockBuffer(buf, BUFFER_LOCK_EXCLUSIVE);
+
+        GenericXLogState *state = GenericXLogStart(index);
```
Ngalstyan4, line 70: This is also an issue in the current version of the `StoreExternalIndex` function.

We need to make sure committed database edits never leave the database at an undefined state.
If some failure happens in one of the iterations of this `for` loop, some blockmap pages would be successfully saved on WAL but the rest will not be, because of the failure. 
This will cause the index be permanently corrupt and unrecoverable.

We need to do `GenericXLogStart` and `GenericXLogFinish` such that when `GenericXLogFinish` is called, the database is always in a valid state. We can merge this as is and fix this separately, as other code that is currently unchanged in StoreExternalIndexBlockMapGroup is also affected by this.

src/hnsw/external_index.c
```diff
@@ -596,13 +612,24 @@ void ldb_wal_retriever_area_free()
 #endif
 }
 
+BlockNumber getBlockMapPageBlockNumber(HnswIndexHeaderPage* hdr, int id) {
+    assert(id >= 0);
+    // Trust me, I'm an engineer!
+    id = id / HNSW_BLOCKMAP_BLOCKS_PER_PAGE + 1;
+    int k;
+    for (k = 0; id >= (1 << k); ++k) {}
+    return hdr->blockmap_page_group_index[k - 1] + (id - (1 << (k - 1)));
```
Ngalstyan4, line 621: դզե՛ց Ճ
Այ սրա համար էի էս տվել դու անեիր

src/hnsw/external_index.c
Ngalstyan4: and the constant should be `HNSW_MAX_BLOCKMAP_GROUPS` right?

Ngalstyan4 (commented):
Well done!
Excited to have this in main soon and start benchmarking. Will continue looking through in the morning!

src/hnsw/external_index.c
davkhech: yep

src/hnsw/external_index.c
davkhech: as we are going to create data blocks afterwards, it is valid

src/hnsw/insert.c
```diff
@@ -156,6 +155,8 @@ bool ldb_aminsert(Relation         index,
     hdr_page = GenericXLogRegisterBuffer(state, hdr_buf, LDB_GENERIC_XLOG_DELTA_IMAGE);
     hdr = (HnswIndexHeaderPage *)PageGetContents(hdr_page);
 
+    HEADER_FOR_EXTERNAL_RETRIEVER = *hdr;
```
davkhech, line 158: during every insert we're reading header page, so makes sense to update retriever's header now

Ngalstyan4:
Thanks! 
#32 is a continuation of this work and the feature will be merged from there.

PR #37, opened 2023-08-09
create index on empty table

davkhech:
when index is crated on empty table there is only header block
when new data is added, we add block map and data pages and populate them

### DIFF
--- a/src/hnsw/external_index.c
+++ b/src/hnsw/external_index.c
@@ -1,4 +1,4 @@
-#include <postgres.h>
+#include "postgres.h"
 
 #include "external_index.h"
 
--- a/src/hnsw/external_index.c
+++ b/src/hnsw/external_index.c
@@ -2,17 +2,17 @@
 
 #include "external_index.h"
 
-#include "cache.h"
-#include "insert.h"
-#include "usearch.h"
-
 #include <access/generic_xlog.h>  // GenericXLog
 #include <assert.h>
-#include <utils/hsearch.h>
 #include <common/relpath.h>
 #include <pg_config.h>       // BLCKSZ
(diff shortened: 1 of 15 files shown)


test/sql/hnsw_insert_empty_table.sql
```diff
@@ -0,0 +1,11 @@
+CREATE EXTENSION IF NOT EXISTS vector;
+CREATE EXTENSION IF NOT EXISTS lanterndb;
+
+CREATE TABLE small_world (
+    id varchar(3),
+    vector vector(3)
```
Ngalstyan4, line 6: Can you make this test with arrays and drop vector dependency?
And, use the small_world definition from `test_helpers` (just `\ir test_helpers/small_world_arrays.sql`)

test/sql/hnsw_insert_empty_table.sql
```diff
@@ -0,0 +1,11 @@
...
+);
+
+CREATE INDEX ON small_world USING hnsw (vector);
+
+INSERT INTO small_world (id, vector) VALUES ('xxx', '[0,0,0]');
```
Ngalstyan4, line 11: Can you also run searches on an empty table?
Then, also try to insert tuples of different dimensionality.

src/hnsw/external_index.c
```diff
@@ -1,4 +1,3 @@
-
 #include "postgres.h"
```
Ngalstyan4, line 1: Use `<...>` for postgres headers (`#include "postgres.h"`)

Ngalstyan4:
Merged these changes through your newer PR (#41). Thanks for your contribution!

PR #46, opened 2023-08-12
Added dist functions from usearch

### LINKED ISSUE #34: Add a new distance function to LanternDB index
Currently we only support l2-squared distance function ([sql](https://github.com/lanterndata/lanterndb/blob/main/sql/lanterndb.sql#L19-L22), [c](https://github.com/lanterndata/lanterndb/blob/db801adced391cbb57ecde5c333635dc12f94e62/src/hnsw.c#L226-L231) refs)through usearch.
We should add other distance functions such as l1, l2, hamming distance, cosine distance, etc

var77:
## Description
Added new dist functions `cos_dist` and `hamming_dist` which will use usearch implementations you can refer to [this PR](https://github.com/Ngalstyan4/usearch/pull/4) for more information.
`l2sq_dist` function now will also use usearch implementation.

## TODO
- [x] Create new update file for sql
- [x] Detect dist function when used with index, so we can set corresponding metric type on usearch options without creating several operators
- [x] Add generic implementation of dist_func to reject usage outside of index
- [x] Add tests to cover functionality

## Issue
closes #34


test/sql/generic_distance_metric.sql
```diff
@@ -0,0 +1,53 @@
...
+-- and can use cos dist if that was chosen when creating the index
+SELECT * FROM (
+    SELECT id, ROUND(l2sq_dist(vector, array[0,1,0])::numeric, 2) as dist
+    FROM small_world
+    ORDER BY vector <-> array[0,1,0] LIMIT 7
+) v ORDER BY v.dist, v.id;
+
+
+--todo:: the bad case is that, I can still forget and use l2sq_dist instead of <->, in SELECT from table
+-- query and the function will just return null
```
Ngalstyan4, line 47: separate PR

src/hnsw.c
```diff
@@ -189,5 +190,44 @@ Datum l2sq_dist(PG_FUNCTION_ARGS)
...
+    bool          return_null = false;
+    while(m) {
+        if(strncmp(m->name, "PortalContext", 13) == 0) {
+            return_null = true;
+        }
+        m = m->parent;
+    }
+
+    if(return_null) {
+        // todo:: figure out why the Executor Portal is calling this function after the index scan
```
Ngalstyan4, line 227: It would be better to fix this case before we merge this in. Can you factor this out into a separate PR?
for now, `<->` will be the l2sq distance and for all the others people have to explicitly write out `_dist` function name.

Ngalstyan4 (commented):
Looks good!

Ngalstyan4 (commented):
Please add a test file with usage tests

- [ ] Create index with each op class and add some rows
- [ ] Use the wrong distance function in an order by and assure the index is not used (via `explain`)

test/expected/hnsw_dist_func.out
```diff
@@ -0,0 +1,160 @@
...
+   ->  Subquery Scan on v  (cost=51.77..51.86 rows=7 width=48)
+         ->  Limit  (cost=51.77..51.79 rows=7 width=52)
+               ->  Sort  (cost=51.77..54.44 rows=1070 width=52)
+                     Sort Key: ((small_world_cos.vector <-> '{0,1,0}'::real[]))
+                     ->  Seq Scan on small_world_cos  (cost=0.00..31.40 rows=1070 width=52)
+(7 rows)
+
+SELECT * FROM (
+    SELECT id, ROUND( ham_dist(vector, array[0,1,0])::numeric, 2) as dist
+    FROM small_world_ham
```
Ngalstyan4, line 131: Is this query using the index?
can you add an `explain`s for cos and ham to make that clear?

sql/updates/0.0.3--latest.sql
```diff
@@ -0,0 +1,16 @@
+-- functions
+CREATE FUNCTION cos_dist(real[], real[]) RETURNS real
+	AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE FUNCTION ham_dist(real[], real[]) RETURNS real
```
Ngalstyan4, line 5: let's call this `hamming_dist`

src/hnsw.c
```diff
@@ -169,25 +170,62 @@ Datum       hnsw_handler(PG_FUNCTION_ARGS)
...
+{
+    if(a->dim != b->dim) {
+        elog(ERROR, "expected equally sized vectors but got vecors with dimensions %d and %d", a->dim, b->dim);
+    }
+
+    return usearch_dist(a->x, b->x, metric_kind, a->dim, usearch_scalar_f32_k);
 }
 
+PGDLLEXPORT PG_FUNCTION_INFO_V1(ldb_generic_dist);
+Datum       ldb_generic_dist(PG_FUNCTION_ARGS) { PG_RETURN_NULL(); }
```
Ngalstyan4, line 198: Is this ever called? can this return an error instead to make sure it is never called directly?

src/hnsw/options.c
```diff
@@ -112,11 +144,25 @@ bytea *ldb_amoptions(Datum reloptions, bool validate)
 #endif
 }
 
+void executor_hook(QueryDesc *queryDesc, int eflags)
+{
+    if(original_executor_start_hook) {
+        original_executor_start_hook(queryDesc, eflags);
+    }
+
+    if(CheckOperatorUsage(queryDesc->sourceText)) {
```
Ngalstyan4, line 153: If I get this right, it does not actually parse the query and looks for the relevant string?
So, a query like `SELECT 'ORDER BY vector <-> [0,0,0]'` would fail?

Is there a hook that gives you a parsed tree of the query instead of just the string?

test/expected/hnsw.out
```diff
@@ -69,13 +69,13 @@ INFO:  usearch index initialized
  001 | 1.00
  010 | 1.00
  100 | 1.00
- 011 | 1.41
- 101 | 1.41
- 110 | 1.41
+ 011 | 2.00
```
Ngalstyan4, line 72: why did the distance change here? was `<->` also not l2sq?

test/expected/hnsw_dist_func.out
```diff
@@ -0,0 +1,224 @@
...
+    ORDER BY vector <-> array[0,1,0] LIMIT 7
+) v ORDER BY v.dist, v.id;
+INFO:  began scanning with 0 keys and 1 orderbys
+INFO:  starting scan with dimensions=3 M=16 efConstruction=128 ef=64
+INFO:  usearch index initialized
+ id  | dist  
+-----+-------
+ 010 |  0.00
+ 011 |  7.00
```
Ngalstyan4, line 190: How is hamming distance calculated here?
I would expect the distance to be 1.00 not 7.00

test/expected/hnsw_dist_func.out
```diff
@@ -0,0 +1,224 @@
...
+         ->  Limit  (cost=0.00..0.53 rows=7 width=52)
+               ->  Index Scan using small_world_ham_vector_idx on small_world_ham  (cost=0.00..81.42 rows=1070 width=52)
+                     Order By: (vector <-> '{0,1,0}'::real[])
+(6 rows)
+
+\set ON_ERROR_STOP off
+-- this should throw error about standalone usage of the operator
+SELECT array[1,2,3] <-> array[3,2,1];
+ERROR:  Operator <-> has no standalone meaning and is reserved for use in vector index lookups only
+-- this should throw error about standalone usage of the operator
```
Ngalstyan4, line 218: Is this because the table has no index?

Ngalstyan4 (approved):
This is amazing!

sql/lanterndb.sql
```diff
@@ -9,25 +9,48 @@ COMMENT ON ACCESS METHOD hnsw IS 'LanternDB vector index access method. Can be c
...
 	COMMUTATOR = '<->'
 );
 
+CREATE OPERATOR <-> (
+	LEFTARG = integer[], RIGHTARG = integer[], PROCEDURE = ldb_generic_dist,
+	COMMUTATOR = '<->'
+);
 
 -- operator classes
 CREATE OPERATOR CLASS ann_l2_ops
```
Ngalstyan4, line 39: ann_l2sq_ops

sql/lanterndb.sql
```diff
@@ -9,25 +9,48 @@ COMMENT ON ACCESS METHOD hnsw IS 'LanternDB vector index access method. Can be c
...
+  DEFAULT FOR TYPE real[] USING hnsw AS
 	OPERATOR 1 <-> (real[], real[]) FOR ORDER BY float_ops,
 	FUNCTION 1 l2sq_dist(real[], real[]);
 
+CREATE OPERATOR CLASS ann_cos_ops
+	FOR TYPE real[] USING hnsw AS
+	OPERATOR 1 <-> (real[], real[]) FOR ORDER BY float_ops,
+	FUNCTION 1 cos_dist(real[], real[]);
+
+CREATE OPERATOR CLASS ann_ham_ops
```
Ngalstyan4, line 49: ann_hamming_ops

sql/lanterndb.sql
Ngalstyan4, line 39: what does `ann` stand for in these operator classes? This is actually something that the clients will type out when creating indexes so we should be thoughtful when naming.
maybe `vec_l2sq_ops`,  `dist_l2sq_ops`?

sql/updates/0.0.3--latest.sql
```diff
@@ -0,0 +1,68 @@
...
+  DEFAULT FOR TYPE real[] USING hnsw AS
+	OPERATOR 1 <-> (real[], real[]) FOR ORDER BY float_ops,
+	FUNCTION 1 l2sq_dist(real[], real[]);
+
+CREATE OPERATOR CLASS ann_cos_ops
+	FOR TYPE real[] USING hnsw AS
+	OPERATOR 1 <-> (real[], real[]) FOR ORDER BY float_ops,
+	FUNCTION 1 cos_dist(real[], real[]);
+
+CREATE OPERATOR CLASS ann_ham_ops
```
Ngalstyan4, line 38: ann_hamming_ops

src/hnsw.h
```diff
@@ -18,4 +20,10 @@
 
 /* Exported functions */
 PGDLLEXPORT void _PG_init(void);
```
Ngalstyan4, line 22: Add `PGDLLEXPORT void _PG_fini(void);
`

src/hnsw/utils.c
```diff
@@ -2,6 +2,9 @@
 
 #include "utils.h"
 
+#include <regex.h>
```
Ngalstyan4, line 5: Does this exist on windows?
It seems to be a POSIX thing.

src/hnsw/utils.c
```diff
@@ -33,7 +36,77 @@ void PopulateUsearchOpts(Relation index, usearch_init_options_t *opts)
...
+        offset += 1;
+    }
+
+    return 0;
+}
+
+int CheckOperatorUsage(const char *query)
+{
+    const char *pattern = "(<->)";
+    const char *orderby_pattern = "order by";
```
Ngalstyan4, line 78: Is query guaranteed to be all lowercase at this point in the executor?

src/hnsw/utils.c
```diff
@@ -33,7 +36,77 @@ void PopulateUsearchOpts(Relation index, usearch_init_options_t *opts)
...
+    regex_t     regex2;
+    regmatch_t  matches[ 1 ];
+    int         reti;
+    int         reti2;
+    int         status = 0;
+
+    reti = regcomp(&regex, pattern, REG_EXTENDED | REG_ICASE);
+    reti2 = regcomp(&regex2, orderby_pattern, REG_EXTENDED | REG_ICASE);
+    if(reti || reti2) {
+        elog(ERROR, "Could not compile regex");
```
Ngalstyan4, line 89: could this ever fail? Do `assert`, if not.

Ngalstyan4: if it could happen, make sure to call `regfree` before erroring

src/hnsw/utils.c
```diff
@@ -33,7 +36,77 @@ void PopulateUsearchOpts(Relation index, usearch_init_options_t *opts)
...
+        return status;
+    }
+
+    // Find all occurrences of the <-> operator
+    int offset = 0;
+    while((reti = regexec(&regex, query + offset, 1, matches, 0)) == 0) {
+        long start = offset + matches[ 0 ].rm_so;
+        long end = offset + matches[ 0 ].rm_eo;
+        char substring[ start - offset ];
+        strncpy(substring, query + offset, start - offset);
```
Ngalstyan4, line 99: Should substring have length start-offset+1 and a null character in the end?

Say offset = 0, start=4
substring is 4 characters long.
strncpy copies start-offset = 4 characters into substring
Substring is pased into regexec shich does not take a length argument so i assume it expects a null terminated string. But what we passed in is not a null terminated string.

test/sql/hnsw_dist_func.sql
```diff
@@ -0,0 +1,133 @@
+CREATE TABLE small_world_l2 (
+    id varchar(3),
+    vector real[]
```
Ngalstyan4, line 3: Can I create a table with `integer[]` column and have l2sq distanced index on it?
Let's have a test for this.

test/sql/hnsw_dist_func.sql
```diff
@@ -0,0 +1,133 @@
...
+);
+
+CREATE TABLE small_world_cos (
+    id varchar(3),
+    vector real[]
+);
+
+CREATE TABLE small_world_ham (
+    id varchar(3),
+    vector integer[]
```
Ngalstyan4, line 13: What happens if I set this to `real[]` ?
does hamming operator refuse to work? does it return the wrong result?
If it is handled, let's add another column to this table (`vector_float`) and also create an index on that in the regression test.

test/sql/hnsw_dist_func.sql
```diff
@@ -0,0 +1,133 @@
...
+);
+
+CREATE TABLE small_world_l2_vec (
+    id varchar(3),
+    vector vector(3)
```
Ngalstyan4, line 18: Would be good to vactor this into a separate file (e.g. `pgvector_compat.sql`)
We will need to then move all the other tests that require pgvector into this file so we could make pgvector dependency optional for running tests as well.

Ngalstyan4 (commented):
Did another pass over this.

sql/lanterndb.sql
var77: it stands for approximate nearest neighbors, but yes I also like `vec_l2sq_ops` one

test/sql/hnsw_dist_func.sql
var77: no, we should add operator class for `integer[]` type in order to work.

test/sql/hnsw_dist_func.sql
var77: no, it will not work now, we should add operator class for `real[]` type for hamming distance in order to work.

src/hnsw/utils.c
var77: You're right, it may make overflows, I will ensure that substring is null terminated before passing to regexec

src/hnsw/utils.c
var77: I assume this to not fail, but I read the examples in manual and in case of error it was not calling `regfree`. 
But I think I should explicitly check there if one was compiled successfully free it before throwing, as there might be case when one fails to compile but the other one compiles successfully

Issue #43, opened 2023-08-11
Move to `pg_regress` for regression testing


Ngalstyan4:
Currently, for testing we simply run `diff` between expected output and the output generated on the instance being tested.

We should move to using [`pg_regress`](https://wiki.postgresql.org/wiki/Regression_test_authoring).
`pg_regress` is a binary available from postgres server dev packages. It is usually not added to `$PATH`, but is available under
`$(pg_config --pkglibdir)/pgxs/src/test/regress/pg_regress`. `pg_regress` calls `diff` under the hood, but advantages of `pg_regress` are:

 The advantages of `pg_regress` are:
- Better edge-case handling (empty output, database unreachable, etc)
- Ability to ignore select failing tests
- Running tests in parallel
- Cross-platform (e.g. see #40 for an example platform issue, when using `diff` naively)

PR #50, opened 2023-08-12
Run tests with pg_regress

var77:
## Description
Run tests with `pg_regress` instead of doing it manually

## TODO

- [x] Remove logs and comments from test results
- [x] Make tests work with filters

## Issue
[#43](https://github.com/lanterndata/lanterndb/issues/43)


test/sql/debug_helpers.sql
```diff
@@ -1,8 +1,9 @@
-
 CREATE EXTENSION IF NOT EXISTS vector;
 CREATE EXTENSION IF NOT EXISTS lanterndb;
```
Ngalstyan4, line 2: I think `pg_regress` has an option to create the extension when it is invoked so we do not have to add these in all files.

test/sql/debug_helpers.sql
```diff
@@ -1,8 +1,9 @@
...
+
 -- initially, before we create an index, the variable does not exist
-\ir test_helpers/small_world.sql
+\ir sql/test_helpers/small_world.sql
+\ir sql/test_helpers/common.sql
```
Ngalstyan4, line 6: Where is `common.sql`? What is it for?

Can you make sure this works with `livedebug.py`?
I think currently there is an issue that we cannot `livedebug` any test that `\ir`s another file because the file path is not resolved correctly. 
We should make sure `livedebug` works for this case or think of a different way to include sql test dependencies.

test/sql/hnsw_type_checks.sql
```diff
@@ -1,6 +1,7 @@
 CREATE EXTENSION IF NOT EXISTS lanterndb;
 
 \set ON_ERROR_STOP off
+DROP TABLE IF EXISTS small_world;
```
Ngalstyan4, line 4: is this needed? doesn't `pg_regress` drop the database between tests?

Ngalstyan4 (commented):
Looks good!

test/sql/hnsw_type_checks.sql
var77: unfortunately I couldn't find any way, so `pg_regress` would drop the database after each case, one option is to iterate over test files and execute `pg_regress` for each one of them.

test/sql/debug_helpers.sql
var77: The `common.sql` was previously executed from `run_all_tests.sh` file before each test case.
Yes currently this is not going to work for `livedebug` script, I'll try to be sure that test suits works with livedebug

test/sql/debug_helpers.sql
var77: It has an option `--load-extension=` , but if I was loading the extension like this, I was getting error that prop `hnsw.init_k` does not exists. I will try to debug and make this work.

scripts/livedebug.py
```diff
@@ -95,6 +100,12 @@ def livedebug():
     # 5. load sql file in psql pane if a file was provided
     time.sleep(1.3)
     if args.file:
+        # as our sql imports are done vie \ir sq/file.sql
+        # we should be in /tests directory so the imports
+        # will resolve the file paths correctly
+        file_dir = '/'.join(args.file.split('/')[:-2])
+        python_pane.send_keys(f"\cd {file_dir}")
```
Ngalstyan4, line 107: `split_window` above has a `start_directory` [option](https://libtmux.git-pull.com/reference/windows.html#libtmux.Window.split_window). Probably cleaner to use that instead of this

scripts/run_all_tests.sh
```diff
@@ -1,13 +1,13 @@
 #!/bin/bash
-# bash strict mode
-set -euo pipefail
```
Ngalstyan4, line 3: Did this cause issues?
It forces the bash script to fail if some component fails and that is generally the sane thing to do.
See [bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/).

scripts/run_all_tests.sh
```diff
@@ -1,13 +1,13 @@
...
 IFS=$'\n\t'
 
 TESTDB=testdb
 PSQL=psql
 TMP_ROOT=/tmp/lanterndb
 TMP_OUTDIR=$TMP_ROOT/tmp_output
 FILTER="${FILTER:-}"
+# this will be used by pg_regress while making diff file
+export PG_REGRESS_DIFF_OPTS=-u
 # read the first command line argument
```
Ngalstyan4, line 11: outdated comment? delete, if so

test/expected/wiki.out
```diff
@@ -1,34 +1,23 @@
...
+\qecho
 
+\set ON_ERROR_STOP on
 SET statement_timeout = 0;
-SET
 SET lock_timeout = 0;
-SET
 SET idle_in_transaction_session_timeout = 0;
-SET
 SET client_encoding = 'UTF8';
```
Ngalstyan4, line 7: The SQL corresponding to all these `SET` logs is unnecessary and does not exist elsewhere.
originally, `wiki.sql` was the result of a `pg_dump`, that is how all of this ended up here. But We should delete all that (now or later) so the wiki test looks like the rest of the tests.

test/test_runner.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+# Get current test file name
+TESTFILE_NAME=${PGAPPNAME##pg_regress/}
+# Set different name for each test database
+# As pg_regress does not support cleaning db after each test
+TEST_CASE_DB="ldb_${TESTFILE_NAME}"
+
+# Drop db after each test on exit signal
+function drop_db {
```
Ngalstyan4, line 10: `pg_regress` has the option of running tests in parallel. How does that option interact with this dropping logic?

test/test_runner.sh
```diff
@@ -0,0 +1,36 @@
...
+  cat <<EOF | psql "$@" -U postgres -d postgres -v ECHO=none >/dev/null 2>&1
+    SET client_min_messages=ERROR;
+    DROP DATABASE "${TEST_CASE_DB}";
+EOF
+}
+
+trap drop_db EXIT
+
+# install lanterndb extension
+psql "$@" -U postgres -d postgres -v ECHO=none -q -c "DROP DATABASE IF EXISTS ${TEST_CASE_DB};" 2>/dev/null
```
Ngalstyan4, line 20: Is there a way to not assume `postgres` user? Is the postgres user guaranteed to exist and be accessible via the current shell?
I just want to make sure tests run in the largest number of cases, regardless of people's postgres setup.

test/test_runner.sh
```diff
@@ -0,0 +1,36 @@
...
+psql "$@" -U postgres -d postgres -v ECHO=none -q -c "DROP DATABASE IF EXISTS ${TEST_CASE_DB};" 2>/dev/null
+psql "$@" -U postgres -d postgres -v ECHO=none -q -c "CREATE DATABASE ${TEST_CASE_DB};" 2>/dev/null
```
Ngalstyan4, line 22: From [postgres docs](https://www.postgresql.org/docs/current/regress-run.html)
> When using installcheck mode, these tests will create and destroy test databases whose names include regression, for example pl_regression or contrib_regression. Beware of using installcheck mode with an installation that has any non-test databases named that way.

It seems the upstream approach is to create and restroy databases hosting the tests, in stead of using an existing database. We should follow the same convention.

scripts/livedebug.py
```diff
@@ -46,13 +49,15 @@ def livedebug():
     new_pane = python_pane.split_window(False, True, ".", 50)
 
     if args.resetdb:
-        res = subprocess.run(f"psql postgres -c 'DROP DATABASE IF EXISTS {args.db};'", shell=True)
-        res = subprocess.run(f"psql postgres -c 'CREATE DATABASE {args.db};'", shell=True)
+        res = subprocess.run(f"psql postgres -U {args.user} -c 'DROP DATABASE IF EXISTS {args.db};'", shell=True)
+        res = subprocess.run(f"psql postgres -U {args.user} -c 'CREATE DATABASE {args.db};'", shell=True)
+        res = subprocess.run(f"psql postgres -U {args.user} -c 'CREATE EXTENSION vector;' -d {args.db}", shell=True)
```
Ngalstyan4, line 54: Can this be optional (behind a flag?)

Ngalstyan4 (commented):
Looks god!

scripts/run_all_tests.sh
var77: I removed it, so I can use `trap` command [here](https://github.com/lanterndata/lanterndb/pull/50/files/050358687cd6d80e233e34eb502f7c024f66f9df#diff-37a6113e1968944c1e8d34be76b1ce800af332c777659556edb55904af465122L69)

test/test_runner.sh
var77: Yes currently all the tests are being run in parallel, and each test case is creating a database for it's own with the name like `ldb_hnsw` or `ldb_hnsw_inserts` like this.

test/test_runner.sh
var77: Yes I will take the user from arguments

test/test_runner.sh
var77: Yes actually we are creating and destroying the database after each test case, I just put `IF EXISTS` here, so if the script will be terminated in the middle it won't cause any issues in next run.

scripts/livedebug.py
var77: Yes I will put it behind a flag

test/expected/wiki.out
var77: I will modify the test in this PR

PR #51, opened 2023-08-14
No global vars

Ngalstyan4:
- [x] Update  usearch API to take a postgres context and pass it to all external retrievers(the context is initialized and passed to usearch iv usearch_init, and it is passed back to postgress in all retriever calls).
- [x] Move `extra_dirtied` page tracking to per-index-operation context
- [x] Move dirtied header to per-index-operation context and get rid of that last global variable
- [ ] ~~Rethink variables names and, generally, cleanup `external_index.c`~~ (cleaned up a little, needs more work but let's do that later)

Incorporates the commit from #67 and closes #67

### DIFF
--- a/scripts/run_all_tests.sh
+++ b/scripts/run_all_tests.sh
@@ -6,7 +6,18 @@ PSQL=psql
 TMP_ROOT=/tmp/lanterndb
 TMP_OUTDIR=$TMP_ROOT/tmp_output
 FILTER="${FILTER:-}"
-DB_USER="${DB_USER:-}"
+# $USER is not set in docker containers, so use whoami
+DEFAULT_USER=$(whoami)
+
+# typically default user is root in a docker container
+# and in those cases postgres is the user with appropriate permissions
+# to the database
+if [ "$DEFAULT_USER" == "root" ]
+then
+    DEFAULT_USER="postgres"
+fi
+
+DB_USER="${DB_USER:-$DEFAULT_USER}"
 # this will be used by pg_regress while making diff file
 export PG_REGRESS_DIFF_OPTS=-u
 
@@ -67,4 +78,4 @@ function print_diff {
 
 trap print_diff ERR
 
-$(pg_config --pkglibdir)/pgxs/src/test/regress/pg_regress --schedule=$SCHEDULE --outputdir=$TMP_OUTDIR --launcher=./test_runner.sh
+DB_USER=$DB_USER $(pg_config --pkglibdir)/pgxs/src/test/regress/pg_regress --user=$DB_USER --schedule=$SCHEDULE --outputdir=$TMP_OUTDIR --launcher=./test_runner.sh
--- a/src/hnsw/external_index.h
+++ b/src/hnsw/external_index.h
@@ -35,7 +35,6 @@ typedef struct HnswIndexHeaderPage
     uint32      vector_dim;
     uint32      num_vectors;
     BlockNumber last_data_block;
-    uint32      blockno_index_start;
     char        usearch_header[ 64 ];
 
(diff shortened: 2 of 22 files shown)


src/hnsw/external_index.h
```diff
@@ -22,20 +22,20 @@
...
-    uint32 blockno_index_start;
-    // todo:: get rid of this
-    uint32 num_blocks;
-    char   usearch_header[ 64 ];
+    uint32      magicNumber;
+    uint32      version;
+    uint32      vector_dim;
+    uint32      num_vectors;
+    BlockNumber last_data_block;
+    uint32      blockno_index_start;
```
davkhech, line 37: I think we can get rid of this one as well. Maybe in separate PR though

src/hnsw/extra_dirtied.c
```diff
@@ -0,0 +1,78 @@
...
+    assert(ed != NULL);
+    ed->index_rel = index_rel;
+    ed->EXTRA_DIRTIED_BLOCKNO = palloc0(sizeof(BlockNumber) * LDB_HNSW_INSERT_MAX_EXTRA_DIRTIED_BUFS);
+    ed->EXTRA_DIRTIED_BUF = palloc0(sizeof(Buffer) * LDB_HNSW_INSERT_MAX_EXTRA_DIRTIED_BUFS);
+    ed->EXTRA_DIRTIED_PAGE = palloc0(sizeof(Page) * LDB_HNSW_INSERT_MAX_EXTRA_DIRTIED_BUFS);
+    ed->EXTRA_DIRTIED_SIZE = 0;
+    return ed;
+}
+
+void extra_dirtied_add(ExtraDirtiedBufs *ed, BlockNumber blockno, Buffer buf, Page page)
```
davkhech, line 19: maybe not good suggestion, but I'd rather pass the context and then get the buffers from the context here

src/hnsw/external_index.h
```diff
@@ -58,8 +58,6 @@ typedef struct HnswIndexTuple
     char   node[ FLEXIBLE_ARRAY_MEMBER ];
 } HnswIndexTuple;
 
-#define HNSW_BLOCKMAP_BLOCKS_PER_PAGE 2000
```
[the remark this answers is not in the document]
Ngalstyan4: we still have it. just moved it to the top of the file.

davkhech (approved):
LGTM

scripts/run_all_tests.sh
```diff
@@ -6,7 +6,7 @@ PSQL=psql
 TMP_ROOT=/tmp/lanterndb
 TMP_OUTDIR=$TMP_ROOT/tmp_output
 FILTER="${FILTER:-}"
-DB_USER="${DB_USER:-}"
+DB_USER="${DB_USER:-postgres}"
```
var77, line 9: I think we can set default to os user 
```suggestion
DB_USER="${DB_USER:-$USER}"
```

also this part can be removed https://github.com/lanterndata/lanterndb/blob/main/test/test_runner.sh#L10-L21

var77 (approved):

test/sql/results/..out.diff
davkhech, line 1: remove empty file

test/sql/results/debug_helpers.sql.out.diff
davkhech, line 1: remove empty file

scripts/run_all_tests.sh
Ngalstyan4: Good catch! fixed, and also made sure to use "postgres" in place of "root" when root is the user.
This is useful for when we are in docker containers

PR #53, opened 2023-08-14
Added autogeneration for lanterdb.control file from CMake

### LINKED ISSUE #30: Auto-generate lanterndb.control
All the info is already in CMakeLists.txt version so we can generate that file, like we generate others
This will make Release [one step shorter](https://github.com/lanterndata/lanterndb/pull/29/files)

grubdragon:
**Why?**
Only thing that changes with every release is the LANTERNDB_VERSION number, which already needs to be set in CMakeLists.txt
Doing so in the lanterndb.control file manually can be eliminated with this autogeneration added in CmakeLists

**Testing Notes:**
1. Run ```cmake ..``` from the `.\build` folder. It should generate the `lanterndb.control` file in the base directory.

Resolves #30


var77:
Looks good to me!

Ngalstyan4 (commented):
Thanks for your contribution, Aryan!
Can you move the template file to the cmake directory?

grubdragon:
@Ngalstyan4 updated the implementation

CMakeLists.txt
```diff
@@ -139,9 +139,17 @@ add_custom_command(
 
 add_custom_target(phony_always_runs ALL DEPENDS ${CMAKE_BINARY_DIR}/${_script_file})
 
-# todo:: make this generated
+
+# AUTO-GENERATE lanterndb.control file for PostgreSQL
+set(CONTROL_TEMPLATE "${CMAKE_SOURCE_DIR}/cmake/lanterndb.control.template")
+set(CONTROL_OUTPUT "${CMAKE_SOURCE_DIR}/lanterndb.control")
```
Ngalstyan4, line 145: sorry, just noticed this. Can you put the generated file in the build directory?
'./build' is in gitignore and nothing there will be tracked. but since `cmake` now puts `lanterndb.control` in the source directory, there will be an untracked file under git after running cmake.

CMakeLists.txt
```diff
@@ -139,9 +139,17 @@ add_custom_command(
...
+
+configure_file(${CONTROL_TEMPLATE} ${CONTROL_OUTPUT})
+install(FILES ${CONTROL_OUTPUT} DESTINATION ${PostgreSQL_EXTENSION_DIR})
+
 set(_control_file "lanterndb.control")
```
Ngalstyan4, line 150: This will need updating accordingly, once `lanterndb.control` moves to the build directory

CMakeLists.txt
aryanagal: Makes sense, pushing the change!

grubdragon: Good idea, we don't want this file to be showing up in our PRs constantly!

CMakeLists.txt
grubdragon: Fixed

grubdragon:
@Ngalstyan4

PR #62, opened 2023-08-19
Fixed cache memory allocation bug

var77:
We were allocating cache memory in  `PortalContext` and it seems that after we free that memory postgres was also trying to free it when cleaning `PortalContext`. 
Simple solution is to allocate the memory in `CacheMemoryContext` which lives longer, this change will not cause memory leaks, as we are destroying the cache memory after end scan and insert operations.

Relates to #61


Ngalstyan4:
Good catch! Pulled your commit to #61 , Lets discuss there

PR #70, opened 2023-08-21
cleanup retrievers

davkhech:
dedublicating the part when we fine the data block number in retrievers

### DIFF
--- a/src/hnsw/external_index.c
+++ b/src/hnsw/external_index.c
@@ -516,13 +516,17 @@ static BlockNumber getBlockMapPageBlockNumber(uint32 *blockmap_page_group_index,
 
 BlockNumber getDataBlockNumber(RetrieverCtx *ctx, int id, bool add_to_extra_dirtied)
 {
-    Cache      *cache = &ctx->block_numbers_cache;
-    uint32     *blockmap_group_index = ctx->header_page_under_wal != NULL
-                                           ? ctx->header_page_under_wal->blockmap_page_group_index
-                                           : ctx->blockmap_page_group_index_cache;
-    BlockNumber blockmapno = getBlockMapPageBlockNumber(blockmap_group_index, id);
-    Buffer      buf;
-    bool        idx_pagemap_prelocked = false;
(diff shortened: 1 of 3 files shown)


src/hnsw/external_index.c
```diff
@@ -547,47 +541,61 @@ void *ldb_wal_index_node_retriever(void *ctxp, int id)
...
-            LockBuffer(buf, BUFFER_LOCK_SHARE);
-            page = BufferGetPage(buf);
-        } else {
-            idx_pagemap_prelocked = true;
+        return blockno_from_cache;
+    }
+
+    // it is necessary to first check the extra dirtied pages for the blockmap page, in case we are in the
+    // middle of an insert and the insert operation has the block we need under a lock
+    // if this is th elast page and blocknos is not filled up, only read the part that is filled up
```
Ngalstyan4, line 549: I think this comment (```/if this is...```) is out of sync with the code. and the todo below is no longer true.

src/hnsw/external_index.c
```diff
@@ -547,47 +541,61 @@ void *ldb_wal_index_node_retriever(void *ctxp, int id)
...
+    // todo:: when we are adding vectors, this does not take into account that the blockmap now has soft-length
+    // of num_vectors + 1
+    Page page = extra_dirtied_get(ctx->extra_dirted, blockmapno, NULL);
+    if(page == NULL) {
+        buf = ReadBufferExtended(ctx->index_rel, MAIN_FORKNUM, blockmapno, RBM_NORMAL, NULL);
+        LockBuffer(buf, BUFFER_LOCK_SHARE);
+        page = BufferGetPage(buf);
+        if(add_to_extra_dirtied) {
+            extra_dirtied_add(ctx->extra_dirted, blockmapno, buf, page);
```
Ngalstyan4, line 558: Wait, we CANNOT lock the page with `BUFFER_LOCK_SHARE` and end up dirtying it. is something off here?

davkhech: I think you're right. During reads we should acquire share lock, during writes exclusive lock.

PR #72, opened 2023-08-21
Improve dev workflow and add SQL code formatter

dqii:


.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
+    "recommendations": [
+        "twxs.cmake",
+        "bradymholt.pgformatter"
```
Ngalstyan4, line 4: Let's not commit commit custom editor configs in here. We will all have custom configs and we can each maintain those.
for pgformatter, we should have a `make` target also.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "editor.formatOnSave": true
```
Ngalstyan4, line 2: Let's not commit this file

CMakeLists.txt
```diff
@@ -18,66 +26,124 @@ if(POLICY CMP0077)
   cmake_policy(SET CMP0077 NEW)
 endif()
 
-# OPTIONS
+# -----------------------------------------------------------------------------
+# Options
+# -----------------------------------------------------------------------------
+
 option(DEV "Developer mode: provide code formatting, get postgres source, etc." OFF)
+
```
Ngalstyan4, line 34: no empty lines between options

sql/lanterndb.sql
```diff
@@ -1,77 +1,114 @@
...
-	
-CREATE FUNCTION ldb_generic_dist(integer[], integer[]) RETURNS real
-	AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
-	
-CREATE FUNCTION l2sq_dist(real[], real[]) RETURNS real
-	AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+CREATE FUNCTION ldb_generic_dist (real[], real[])
+    RETURNS real
+    AS 'MODULE_PATHNAME'
+    LANGUAGE C
```
Ngalstyan4, line 15: Do we have the option of turning off `pgFormatter` formatting for parts of the file (e.g. via a magic-comment.
e.g. in clang-format, any code I write within the magic comments below will not be formatted
```
// clang-format off
[CODE HERE]
// clang-format on
```
We should use a formatter that has such an option and should use such an option for sql statements that are very similar. This allows to more easily tell what is different between similar SQL statements.


For example, in the above, the two `ldb_generic_dist`, and `l2sq_dist` only differ in their argument types and that is easier to quickly verify visually when we have long "unformatted" sql lines.

sql/updates/0.0.1--0.0.2.sql
```diff
@@ -1,18 +1,27 @@
...
+
 DROP ACCESS METHOD IF EXISTS embedding;
+
 DROP FUNCTION IF EXISTS hnsw_handler;
 
-CREATE FUNCTION hnsw_handler(internal) RETURNS index_am_handler
-	AS 'MODULE_PATHNAME' LANGUAGE C;
+CREATE FUNCTION hnsw_handler (internal)
+    RETURNS index_am_handler
+    AS 'MODULE_PATHNAME'
```
Ngalstyan4, line 11: I have a weak preference of not touching old update files. These are supposed to be immutable after the corresponding release. Formatting does not change the logic so not a big deal

.vscode/settings.json
dqii: Re: removing this and the extensions.json --- understood, will remove these.

Is there a reason to include `.vscode/c_cpp_properties.json`? If not I can remove that from repo and add `.vscode` to `.gitignore`.

I can add these all to future documentation as suggestions for dev workflow?

.vscode/extensions.json
dqii: > for pgformatter, we should have a make target also.

I added a make target! It's called `format-sql`. Should I mention it in future docs?

The flow I plan to follow is that CI/CD will check if relevant files are mis-formatted, and throw an error if so. The user will need to format these files, whether through running `make format-sql`, VSCode auto-formatter, etc.

sql/lanterndb.sql
dqii:
```
 -p | --placeholder RE : set regex to find code that must not be changed.
```
We could use this to achieve that.

.vscode/settings.json
Ngalstyan4: Let's remove that and the `.vscode` folder also

sql/lanterndb.sql
Ngalstyan4: Ok, great! 
Let's do that for these kind of repeating statements.

.vscode/settings.json
dqii: Sounds good. I saw timescale has `/.vscode` in `.gitignore`

dqii:
Update

pgFormatter doesn't support magic comment ignores well. (It keeps adding a space to the next line, for example). In addition, it doesn't support CTEs well either (multiple Github issues about this).

I tried out pglast, which is based on libpg_query, which uses Postgres's parser. This handles all the Postgres SQL and pl/pgSQL corner cases nicely, but unfortunately doesn't handle the psql metacommands well (e.g., \set \eqcho \copy)

PR #76, opened 2023-08-22
Enable dynamic undefined symbols in shared libraries on Apple

Ngalstyan4:
…

The default in new mac systems seems to be that undefined symbols are allowed in shared libraries. This is the default in linux systems as well.

So far we assumed all mac systems allow undefined symbols in shared libraries. But that is not the case and old linkers need the extra flag


smmuirhead100:
Was having trouble with installation, but this fixed it. Thanks!

PR #78, opened 2023-08-24
remove hard limit on the number of items returned when an index is queried

### LINKED ISSUE #74: Remove arbitrary upper limit for number of returned results from hnsw index
Currently our index returns at most 50 elements (see [here](https://github.com/lanterndata/lanterndb/blob/main/src/hnsw/scan.c#L169-L171))
So `SELECT .. LIMIT 50` and `SELECT .. LIMIT 100` will both return 50 results.

The correct behaviour is to return the number of results the query asked for.
In the index scanning routine we do not know how many results the user has asked for. So, we should initially retrieve a fixed number of results from the index (e.g. 50, this value should be configured according to [`init_k`](https://github.com/search?q=repo%3Alanterndata%2Flanterndb%20init_k&type=code)) and then retrieve more (`2*init_k`, `4*init_k`...) until scanning is over.

ezra-varady:
This should address #74 It includes a regression test where `...LIMIT 10` (the new default) and `...LIMIT 50` are tested.

Resolves #74


test/sql/hnsw_index_limit.sql
```diff
@@ -0,0 +1,12 @@
+-- We need a large enough table that postgres uses the index
+ CREATE TABLE sift_base10k (
+     id SERIAL PRIMARY KEY,
+     v real[128]
+);
+
+\copy sift_base10k (v) FROM '/tmp/lanterndb/vector_datasets/siftsmall_base_arrays.csv' with csv;
+-- build index on an existing table of 10k rows
```
Ngalstyan4, line 8: please add these tests to hnsw.sql instead of creating a new test file

test/sql/hnsw_index_limit.sql
```diff
@@ -0,0 +1,12 @@
...
+CREATE INDEX hnsw_idx ON sift_base10k USING hnsw (v) WITH (M=2, ef_construction=10, ef=4, dims=128);
+SELECT V AS v4444  FROM sift_base10k WHERE id = 4444 \gset
+SELECT * FROM sift_base10k order by v <-> :'v4444' LIMIT 10;
```
Ngalstyan4, line 11: Can you do `SELECT count(*) ...` instead? we really are only testing the count here so lets make the `.out` file less noisy

test/sql/hnsw_index_limit.sql
```diff
@@ -0,0 +1,12 @@
...
+SELECT * FROM sift_base10k order by v <-> :'v4444' LIMIT 50;
```
Ngalstyan4, line 12: Can you think of more edge cases to test here?
E.g. What happens when I am asking for 50 but the table has 30 elements?

src/hnsw/scan.c
```diff
@@ -211,6 +213,31 @@ bool ldb_amgettuple(IndexScanDesc scan, ScanDirection dir)
...
+        Datum           value;
+        Vector         *vec;
+        usearch_error_t error = NULL;
+
+        value = scan->orderByData->sk_argument;
+
+        vec = DatumGetVector(value);
+
+        /* double k and reallocate arrays to account for increased size */
+        k = k * 2;
```
Ngalstyan4, line 227: So you are using the static `k` to know how many items to retrieve next.
Could this be `k = scanstate->count * 2`?

It is very strange to see static variables in this context. I am pretty sure statics will do the wrong thing if two scans happen at the same time. So, if possible, let's get rid of the static variable.

test/sql/hnsw_index_limit.sql
Ngalstyan4, line 12: Can you also test the case when the `init_k` variable is changed?
You may need to add a debug log every time `usearch_search` is called to notice the difference in sql output.

Ngalstyan4 (commented):
Thanks for your contribution, @ezra-varady !
This is great! And is almost ready. See comments below.

src/hnsw/scan.c
ezra-varady: good catch, this was poorly thought out

ezra-varady:
to summarize, 

- I fixed the issue with the static variable. You're absolutely right about it breaking when two queries occur simultaneously. An embarrassing mistake. 
- Moved new tests into hnsw.sql and made corresponding updates to the expected output
- Added debug logging of `init_k` when `usearch_search` is called, as well as tests that modify it 

I still need to add a test for the query being bigger than the table

Ngalstyan4:
Thanks for the prompt fixes, @ezra-varady! 

> I still need to add a test for the query being bigger than the table. 

I am not sure what you want to test here. But if the problem is that the table has more rows than you need, you can place the test in a transaction, drop all the extra rows in the transaction, then run the query, then rollback the transaction.
See an example of this kind of thing [here](https://github.com/lanterndata/lanterndb/blob/c236c4e9d1957da9f97f9809819568da961a0cc1/test/sql/hnsw.sql#L81C6-L95).

Note that you also do not need a large table to force index usage in tests. you can just [disable sequential scan](https://github.com/lanterndata/lanterndb/blob/c236c4e9d1957da9f97f9809819568da961a0cc1/test/sql/hnsw_insert.sql#L7).

ezra-varady:
Thanks, I didn't know that! I reformatted the tests in light of this. I also added a check to avoid a potential extra query I had neglected. I think I had misunderstood your suggestion w.r.t. the additional test. Please let me know if you have other suggestions

src/hnsw/scan.c
```diff
@@ -153,6 +153,7 @@ bool ldb_amgettuple(IndexScanDesc scan, ScanDirection dir)
 {
     HnswScanState *scanstate = (HnswScanState *)scan->opaque;
     ItemPointer    tid;
+    // todo:: fix me. if there is way to know how many we need, use that
```
Ngalstyan4, line 156: unneeded comment or comment in the wrong place

src/hnsw/scan.c
```diff
@@ -191,13 +190,13 @@ bool ldb_amgettuple(IndexScanDesc scan, ScanDirection dir)
...
         if(scanstate->distances == NULL) {
             scanstate->distances = palloc(k * sizeof(float));
-            ;
         }
         if(scanstate->labels == NULL) {
             scanstate->labels = palloc(k * sizeof(usearch_label_t));
         }
 
         // hnsw_search(scanstate->hnsw, vec->x, k, &num_returned, scanstate->distances, scanstate->labels);
+        elog(INFO, "querying index for %d elements", k);
```
Ngalstyan4, line 199: Can you make sure this is only triggered for the tests you wrote and not for all the other tests as well?

Ngalstyan4 (approved):
lgtm!

One small thing:
To avoid the extra logs added to the rest of the regression tests due to the elog here, can you surround the tests that concern the dynamic k-value growth with the following statements?

```
SET client_min_messages TO DEBUG5;
--- YOUR TESTS HERE ---
RESET client_min_messages;
```
Then, if you change your added `elog(INFO,..` code lines to `elog(DEBUG5`, the log message will be printed only in the tests you needed.

As a side note, we should go do this for the rest of the regression tests as well.

ezra-varady:
removed the hanging comment and changed the log level for the tests

PR #87, opened 2023-08-25
Test improvements

dqii:


dqii:
TODO: Redo the out files once bug with REAL[] definition is fixed

var77 (approved):

dqii:
I restructured the tests to avoid duplicate tests, and add the vector tests. I also fixed the discussed bug with DatumGetVector so I could get meaningful results in my out files. I also removed some of the "todos" that were already completed, and added a new "todo" (bug with hamming distance).

var77 (approved):
Great!

Issue #55, opened 2023-08-16
Remove default dimensions when using HNSW index


dqii:
I have a table with column `v REAL[]`. When I tried to create an index I got this error
```
lantern=# CREATE INDEX sift_base10k_index ON sift_base10k USING
        hnsw (v) WITH (
            M=2,
            ef_construction=16,
            ef=10
        );
INFO:  done init usearch index
ERROR:  Wrong number of dimensions: 128 instead of 3 expected
```
It seems like `dims=128` is a required parameter. It would be nice if this could be derived from the data itself.

PR #88, opened 2023-08-25
infer dimensions

ezra-varady:
This should address #55, allowing postgres to automatically decide the dimension index of arrays of `real` and `int`. It  preserves the user's ability to specify a dimension. When `dims` is left as the default lantern will now check the column's dimension and use that instead if it differs. I haven't tested against older versions of postgres, and I'm not familiar with the API, so this may need some additional work to be backwards compatible


ezra-varady:
API appears to have been different in postgres 11, will look into it later this evening

src/hnsw/build.c
```diff
@@ -153,6 +158,56 @@ void CheckHnswIndexDimensions(Relation index, Datum arrayDatum, int dimensions)
...
+#endif
+    HeapTuple  tuple;
+    Snapshot   snapshot;
+    ArrayType *array;
+    Datum      datum;
+    bool       isNull;
+    int        n_items = currDim;
+    int        indexCol;
+
+    // If NumIndexAttrs isn't 1 there's no clear way to infer the dim
```
Ngalstyan4, line 179: Can you clarify in the comment what NumIndexAttrs represents?

src/hnsw/build.c
```diff
@@ -153,6 +158,56 @@ void CheckHnswIndexDimensions(Relation index, Datum arrayDatum, int dimensions)
...
+#endif
+    tuple = heap_getnext(scan, ForwardScanDirection);
+    if(tuple == NULL) {
+        heap_endscan(scan);
+        return n_items;
+    }
+
+    // Get the indexed column out of the row and return it's dimensions
+    datum = heap_getattr(tuple, indexCol, RelationGetDescr(heap), &isNull);
+    if(!isNull) {
```
Ngalstyan4, line 201: So it seems, if the first row in the table has NULL vector and all the ones afterwards have e.g. 3 dimensional real[] arrays, then we fail to infer. This is fair. Let's just add a test for it and document it

src/hnsw/build.c
```diff
@@ -166,6 +221,11 @@ static void InitBuildState(HnswBuildState *buildstate, Relation heap, Relation i
 
     buildstate->dimensions = GetHnswIndexDimensions(index);
 
+    // If a dimension wasn't specified try to infer it
+    // this is unecessary when the actual size is 3 but in practice this shouldn't be too common
+    if(columnType == REAL_ARRAY || columnType == INT_ARRAY)
+        if(buildstate->dimensions == HNSW_DEFAULT_DIMS)
+            buildstate->dimensions = GetArrayLengthFromHeap(heap, indexInfo, buildstate->dimensions);
     /* Require column to have dimensions to be indexed */
     if(buildstate->dimensions < 1) elog(ERROR, "column does not have dimensions");
```
Ngalstyan4, line 230: Is this the error we are going to see if the table is empty? Is that case tested?

Ngalstyan4 (commented):
Good! Left small comments. Should be ready soon

src/hnsw/build.c
```diff
@@ -166,6 +221,11 @@ static void InitBuildState(HnswBuildState *buildstate, Relation heap, Relation i
 
     buildstate->dimensions = GetHnswIndexDimensions(index);
 
+    // If a dimension wasn't specified try to infer it
+    // this is unecessary when the actual size is 3 but in practice this shouldn't be too common
+    if(columnType == REAL_ARRAY || columnType == INT_ARRAY)
+        if(buildstate->dimensions == HNSW_DEFAULT_DIMS)
```
var77, line 227: I think we should set `HNSW_DEFAULT_DIMS` to -1, NULL or something like that, as now if we pass `(dims=3)` on index creation and insert array with 4 dimensions it is going to overwrite the `dims` option.

src/hnsw/build.c
ezra-varady: It would be if `DEFAULT_DIMS` were changed from 3 to -1.  Currently a number of tests create indices on empty tables, so I wasn't sure what the expected behavior was. I preserved the default to maximize compatibility with the existing API, so right now this wouldn't generate an error. If it's alright to throw an error when the table being indexed is empty I can change the default per @var77's suggestion and update the relevant tests to make things more consistent

ezra-varady:
There's actually an issue with `GetHnswIndexDimensions`. If we infer the dimension it won't set the `dims` relopt and it will break scans. Unfortunately afaict while building the index the `rd_options` field is still null, and my various attempts to set it manually don't work, marking as draft for a bit while I add logic to fix this

src/hnsw/build.c
Ngalstyan4: It is all right and the expected behaviour. Just throw a helpful message asking the user to provide number of dimensions for the indexed column

Ngalstyan4:
Hi @ezra-varady,

This might have been a result of some issues on indexes with `real[]` column type. @dqii addressed those at #87 which was just merged to main. Could you rebase and see if the issues you mentioned above persist?

src/hnsw/build.c
```diff
@@ -161,8 +250,11 @@ static void InitBuildState(HnswBuildState *buildstate, Relation heap, Relation i
     buildstate->columnType = GetIndexColumnType(index);
     buildstate->dimensions = GetHnswIndexDimensions(index);
 
+    // If a dimension wasn't specified try to infer it
+    if(buildstate->columnType == REAL_ARRAY || buildstate->columnType == INT_ARRAY)
+        if(buildstate->dimensions < 1) buildstate->dimensions = InferDimension(heap, indexInfo);
```
Ngalstyan4, line 255: We should always use braces `{,}` with `if`s and other basic blocs. 
It is particularly important for cases like these where you have multiple `if`s
 
I think the code above can be clarified as well. 
What happens when `buildstate->columnType` is not `REAL_ARRAY` or `INT_ARRAY`?

src/hnsw/build.c
```diff
@@ -115,14 +120,81 @@ static void BuildCallback(
...
+        // The dimension in options is not set if the dimension is inferred so we need to actually check the key
+        int opt_dim = HnswGetDims(index);
+        if(opt_dim == HNSW_DEFAULT_DIMS) {
+            // If the option's still the default it needs to be updated to match what was inferred
+            // todo: is there a way to do this earlier? (rd_options is null in BuildInit)
+            Relation     heap;
+            HnswOptions *opts;
+            int          attrNum;
+
+            // We know there's one key because we generate an error during inference if multiple keys are specified
```
Ngalstyan4, line 179: Can you add an assert in place of this comment?

src/hnsw/build.c
ezra-varady: Ah you're right, the operators aren't defined, good catch

PR #89, opened 2023-08-25
Switch LINKER_LANGUAGE from C to CXX

grubdragon:
On compiling postgresql and lanterndb from source from scratch, without installing postgresql core, the lanterndb extension compiles and installs, but it is unable to execute `CREATE EXTENSION lanterndb;` with the error `undefined symbol: _ZTVN10__cxxabiv117__class_type_infoE`. This is due to libstc++ not being linked. Switching the LINKER_LANGUAGE to CXX allows the build system to use a C++ linker and solves the issue.

References:
- https://github.com/sensational/sassphp/issues/4
- https://github.com/Iorethan/opencv_pfm/issues/1


grubdragon:
Not the right approach - needs a different soluton. Closing.

PR #90, opened 2023-08-26
[WIP] Get index from file

grubdragon:
Some initial code based on investigation of postgres code


grubdragon:
These changes were reviewed and it was ultimately determined to not be the right way, since using a "stmt" is higher upstream than required for a db. Closing.

PR #93, opened 2023-08-27
Fix range of HNSW's M parameter

grubdragon:


dqii:
Thanks Aryan!

Can these tests be added to hsnw_create instead of creating a separate test file? I updated the tests so that each file is more clear about what it's testing.

Can you only include M=1 and M=128? There's no need for redundant test cases, and M=4 (in range) is already tested in that file.

PR #97, opened 2023-08-29
Only run pgvector test if pgvector is installed

dqii:


test/schedule.txt
```diff
@@ -1,8 +1,10 @@
 # schedule.txt rules:
 # - every test that needs to be run must appear in a 'test:' line
 # - every test that needs to be run and is expected to fail must appear in an 'ignore:' line *before* the corresponding test line
+# - every test that needs to be run iff pgvector is installed appears in a 'pgvector:' line
 # - 'ignore' lines must have exactly one test
 # - 'test' lines may have multiple space-separated tests. All tests in a single 'test' line will be run in parallel
 
 ignore: hnsw_todo
-test: hnsw_config hnsw_correct hnsw_create hnsw_dist_func hnsw_insert hnsw_select hnsw_todo hnsw_vector
+pgvector: hnsw_vector
```
Ngalstyan4, line 9: Can we call this test_pgvector? its semantics is similar to `test:` and very different from `ignore:`

test/test_runner.sh
```diff
@@ -27,9 +27,13 @@ trap drop_db EXIT
 cd sql/
 # install lanterndb extension
 psql "$@" -U ${DB_USER} -d postgres -v ECHO=none -q -c "DROP DATABASE IF EXISTS ${TEST_CASE_DB};" 2>/dev/null
-psql "$@" -U ${DB_USER} -d postgres -v ECHO=none -q -c "DROP DATABASE IF EXISTS ${TEST_CASE_DB};" 2>/dev/null
 psql "$@" -U ${DB_USER} -d postgres -v ECHO=none -q -c "CREATE DATABASE ${TEST_CASE_DB};" 2>/dev/null
-psql "$@" -U ${DB_USER} -d ${TEST_CASE_DB} -v ECHO=none -q -c "SET client_min_messages=error; CREATE EXTENSION vector; CREATE EXTENSION lanterndb;" 2>/dev/null
+VECTOR_AVAILABLE=$(psql -U ${DB_USER} -t -c "SELECT 1 FROM pg_available_extensions WHERE name = 'vector';" -tA)
+if [ "$VECTOR_AVAILABLE" = "1" ]; then
+    psql "$@" -U ${DB_USER} -d ${TEST_CASE_DB} -v ECHO=none -q -c "SET client_min_messages=error; CREATE EXTENSION vector; CREATE EXTENSION lanterndb;" 2>/dev/null
```
Ngalstyan4, line 33: I think we should avoid creating `vector` extension here and should instead assume `sql` files requiring that extension will have a `CREATE EXTENSION vector;` line.
That way, we can make sure that the rest of the tests run without enabling the vector extension.
As is, it would be really easy to have a hidden dependency on vector from other test files without realizing it since we have pgvector installed in ci/cd and on our dev machines.

scripts/run_all_tests.sh
```diff
@@ -46,29 +46,42 @@ then
     echo "Successfully Downloaded all necessary vector test files"
 fi
 
+# Check if pgvector is available
+pgvector_installed=$($PSQL -U $DB_USER -c "SELECT 1 FROM pg_available_extensions WHERE name = 'vector'" -tA)
+
+# Generate schedule.txt
+rm -rf $TMP_OUTDIR/schedule.txt
+if [ -n "$FILTER" ]; then
```
Ngalstyan4, line 54: just curious, what is the difference between old `-z` and new `-n` ?

scripts/run_all_tests.sh
```diff
@@ -46,29 +46,42 @@ then
...
+    if [[ "$pgvector_installed" == "1" ]]; then
+        TEST_FILES=$(cat schedule.txt | grep -E '^(test:|pgvector:)' | sed -e 's/^\(test:\|pgvector:\)//' | tr " " "\n" | sed -e '/^$/d')
```
Ngalstyan4, line 56: We could avoid all the shell wizardry by creating a separate schedule_pgvector.txt schedule with the single test and relying on pg_regress to do the parsing.
This works too though.

Ngalstyan4 (commented):
Looks good! Added small suggestions.

test/test_runner.sh
dqii: I'm happier with that.

scripts/run_all_tests.sh
dqii: -z and -n are pretty much "is null" vs "is not null". I was originally planning to do "if filter is not null then X else if pgvector then Y else Z". I didn't take that route, but this was just a side effect of that.

Ngalstyan4 (approved):

PR #99, opened 2023-08-30
Add ability to load index from file

var77:
## Description
- Added parameter `_experimental_index_path`
- If the `_experimental_index_path` will be defined we will load the index in `BuildIndex` function instead of doing table scan.
- Changed label creation code to take exactly 6 bytes from `ItemPointerData` struct, previously it might fill the last 2 bytes with garbage.
- Added tests with 2 index files to test the functionality

### DIFF
--- a/src/hnsw/build.c
+++ b/src/hnsw/build.c
@@ -315,12 +315,11 @@ static void ScanTable(HnswBuildState *buildstate)
 static void BuildIndex(
     Relation heap, Relation index, IndexInfo *indexInfo, HnswBuildState *buildstate, ForkNumber forkNum)
 {
-    InitBuildState(buildstate, heap, index, indexInfo);
-
     usearch_error_t        error = NULL;
     usearch_init_options_t opts;
     MemSet(&opts, 0, sizeof(opts));
 
+    InitBuildState(buildstate, heap, index, indexInfo);
     opts.dimensions = buildstate->dimensions;
     PopulateUsearchOpts(index, &opts);
 
--- a/src/hnsw/build.c
+++ b/src/hnsw/build.c
@@ -72,14 +72,14 @@ static void AddTupleToUsearchIndex(ItemPointer tid, Datum *values, HnswBuildStat
             pg_unreachable();
     }
 
-        // casting tid structure to a number to be used as value in vector search
-        // tid has info about disk location of this item and is 6 bytes long
+    // casting tid structure to a number to be used as value in vector search
+    // tid has info about disk location of this item and is 6 bytes long
+    usearch_label_t label = GetUsearchLabel(tid);
 #ifdef LANTERN_USE_LIBHNSW
-    if(buildstate->hnsw != NULL) hnsw_add(buildstate->hnsw, vector, *(unsigned long *)tid);
+    if(buildstate->hnsw != NULL) hnsw_add(buildstate->hnsw, vector, label);
 #endif
 #ifdef LANTERN_USE_USEARCH
-    if(buildstate->usearch_index != NULL)
-        usearch_add(buildstate->usearch_index, *(unsigned long *)tid, vector, usearch_scalar, &error);
+    if(buildstate->usearch_index != NULL) usearch_add(buildstate->usearch_index, label, vector, usearch_scalar, &error);
 #endif
     assert(error == NULL);
     buildstate->tuples_indexed++;
@@ -249,6 +249,7 @@ static void InitBuildState(HnswBuildState *buildstate, Relation heap, Relation i
     buildstate->indexInfo = indexInfo;
     buildstate->columnType = GetIndexColumnType(index);
     buildstate->dimensions = GetHnswIndexDimensions(index);
+    buildstate->index_file_path = HnswGetIndexFilePath(index);
 
     // If a dimension wasn't specified try to infer it
     if(buildstate->dimensions < 1) {
@@ -314,28 +315,42 @@ static void ScanTable(HnswBuildState *buildstate)
 static void BuildIndex(
     Relation heap, Relation index, IndexInfo *indexInfo, HnswBuildState *buildstate, ForkNumber forkNum)
 {
+    InitBuildState(buildstate, heap, index, indexInfo);
+
+    usearch_error_t        error = NULL;
     usearch_init_options_t opts;
     MemSet(&opts, 0, sizeof(opts));
-    InitBuildState(buildstate, heap, index, indexInfo);
 
     opts.dimensions = buildstate->dimensions;
     PopulateUsearchOpts(index, &opts);
 
-    usearch_error_t error = NULL;
-    buildstate->hnsw = NULL;
     buildstate->usearch_index = usearch_init(&opts, &error);
-
     elog(INFO, "done init usearch index");
     assert(error == NULL);
 
-    usearch_reserve(buildstate->usearch_index, 1100000, &error);
-    assert(error == NULL);
+    buildstate->hnsw = NULL;
+    if(buildstate->index_file_path) {
+        usearch_load(buildstate->usearch_index, buildstate->index_file_path, &error);
+        assert(error == NULL);
+        elog(INFO, "done loading usearch index");
 
-    UpdateProgress(PROGRESS_CREATEIDX_PHASE, PROGRESS_HNSW_PHASE_IN_MEMORY_INSERT);
-    LanternBench("build hnsw index", ScanTable(buildstate));
+        // todo determine how we can set index rd_options
+        // HnswOptions *opts = (HnswOptions *)index->rd_options;
+        // opts->m = usearch_connectivity(buildstate->usearch_index, &error);
+        // assert(error == NULL);
+        // opts->dims = usearch_dimensions(buildstate->usearch_index, &error);
+        // assert(error == NULL);
+        // todo determine how to get ef construction and ef search params
+    } else {
+        usearch_reserve(buildstate->usearch_index, 1100000, &error);
+        assert(error == NULL);
 
-    elog(INFO, "inserted %ld elements", usearch_size(buildstate->usearch_index, &error));
-    assert(error == NULL);
+        UpdateProgress(PROGRESS_CREATEIDX_PHASE, PROGRESS_HNSW_PHASE_IN_MEMORY_INSERT);
+        LanternBench("build hnsw index", ScanTable(buildstate));
+
+        elog(INFO, "inserted %ld elements", usearch_size(buildstate->usearch_index, &error));
+        assert(error == NULL);
+    }
 
     char *result_buf = NULL;
     usearch_save(buildstate->usearch_index, NULL, &result_buf, &error);
--- a/src/hnsw/options.c
+++ b/src/hnsw/options.c
@@ -12,6 +12,12 @@
 #include <utils/rel.h>  // RelationData
 #include <utils/syscache.h>
 
+#ifdef _WIN32
+#define access _access
+#else
+#include <unistd.h>
+#endif
+
 // We import this header file
 // to access the op class support function pointers
 #include "../hnsw.h"
@@ -52,6 +58,17 @@ int HnswGetEf(Relation index)
     HnswOptions *opts = (HnswOptions *)index->rd_options;
     if(opts) return opts->ef;
     return HNSW_DEFAULT_EF;
+}
+
+char *HnswGetIndexFilePath(Relation index)
+{
+    HnswOptions *opts = (HnswOptions *)index->rd_options;
+    if(!opts) return NULL;
+    if(!opts->experimantal_index_path_offset) {
+        return NULL;
+    }
+
+    return (char *)opts + opts->experimantal_index_path_offset;
 }
 
 usearch_metric_kind_t HnswGetMetricKind(Relation index)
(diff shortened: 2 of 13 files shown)


src/hnsw/build.c
```diff
@@ -221,28 +222,42 @@ static void ScanTable(HnswBuildState *buildstate)
...
+    if(buildstate->index_file_path) {
+        usearch_load(buildstate->usearch_index, buildstate->index_file_path, &error);
+        assert(error == NULL);
+        elog(INFO, "done loading usearch index");
+
+        // todo determine how we can set index rd_options
+        // HnswOptions *opts = (HnswOptions *)index->rd_options;
+        // opts->m = usearch_connectivity(buildstate->usearch_index, &error);
+        // assert(error == NULL);
+        // opts->dims = usearch_dimensions(buildstate->usearch_index, &error);
```
Ngalstyan4, line 248: You can probably add all these (usearch_connectivitydims ) to existing [`usearch_metadata`](https://github.com/Ngalstyan4/usearch/blob/c9454d98e8af2402a1bd874db182dd8530d44350/c/lib.cpp#L172-L181)

src/hnsw/build.c
```diff
@@ -314,28 +315,42 @@ static void ScanTable(HnswBuildState *buildstate)
 static void BuildIndex(
     Relation heap, Relation index, IndexInfo *indexInfo, HnswBuildState *buildstate, ForkNumber forkNum)
 {
+    InitBuildState(buildstate, heap, index, indexInfo);
```
Ngalstyan4, line 318: Note that per convention all local variable declarations should come before any statements;

src/hnsw/build.c
Ngalstyan4: @var77 I assume you left these comments in to continue working on these in a separate PR?

src/hnsw/options.c
```diff
@@ -12,6 +12,12 @@
 #include <utils/rel.h>  // RelationData
 #include <utils/syscache.h>
 
+#ifdef _WIN32
```
Ngalstyan4, line 15: What is this about? did you test it on windows?!

src/hnsw/options.c
```diff
@@ -108,21 +125,28 @@ static void HnswAlgorithmParamValidator(const char *value)
     }
 }
 
+static void IndexFileParamValidator(const char *value)
+{
+    if(value == NULL) return;
+
+    if(access(value, F_OK) != 0) {
+        ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("Invalid index file path ")));
```
Ngalstyan4, line 133: Can we trigger this case in a test?

src/hnsw/options.c
```diff
@@ -108,21 +125,28 @@ static void HnswAlgorithmParamValidator(const char *value)
...
-                                           //  or maybe it just does not put anything and does not even reset?
-                                           {"alg", RELOPT_TYPE_STRING, offsetof(HnswOptions, alg)}};
+    static const relopt_parse_elt tab[]
+        = {{"dims", RELOPT_TYPE_INT, offsetof(HnswOptions, dims)},
+           {"element_limit", RELOPT_TYPE_INT, offsetof(HnswOptions, element_limit)},
+           {"m", RELOPT_TYPE_INT, offsetof(HnswOptions, m)},
+           {"ef_construction", RELOPT_TYPE_INT, offsetof(HnswOptions, ef_construction)},
+           {"ef", RELOPT_TYPE_INT, offsetof(HnswOptions, ef)},
+           {"_experimental_index_path", RELOPT_TYPE_STRING, offsetof(HnswOptions, experimantal_index_path_offset)},
+           {"alg", RELOPT_TYPE_STRING, offsetof(HnswOptions, alg_offset)}};
```
Ngalstyan4, line 149: While you are here, please get rid of the "alg" parameter.

test/sql/hnsw_index_from_file.sql
```diff
@@ -0,0 +1,30 @@
...
+------------------------------------------------------------------------------
+
+-- Index files were created with ldb-create-index program which source is under https://github.com/lanterndata/lanterndb_extras/
+-- We have exported index files for sift1k dataset for cosine and l2sq distances
+-- With the following params m=16 ef=64 ef_construction=128 dims=128
+-- Validate that index creation works with a small number of vectors
+\ir utils/sift1k_array.sql
+
+-- Validate that creating an index from file works
+CREATE INDEX hnsw_l2_index ON sift_base1k USING hnsw (v dist_l2sq_ops) WITH (dims=128, M=16, ef=64, ef_construction=128, _experimental_index_path='/tmp/lanterndb/files/index-sift1k-l2.usearch');
```
Ngalstyan4, line 12: Is it required to pass dims, M, ef,... parameters? What happens if you do not?

Ngalstyan4: Oh, I guess it is required now but we will drop the requirement once we store that info in the index header page?

test/sql/hnsw_index_from_file.sql
```diff
@@ -0,0 +1,30 @@
...
+SELECT * FROM ldb_get_indexes('sift_base1k');
+
+SET enable_seqscan = false;
+
+SELECT v AS v777 FROM sift_base1k WHERE id = 777 \gset
+EXPLAIN (COSTS FALSE) SELECT ROUND(l2sq_dist(v, :'v777')::numeric, 2) FROM sift_base1k order by v <-> :'v777' LIMIT 10;
+SELECT ROUND(l2sq_dist(v, :'v777')::numeric, 2) FROM sift_base1k order by v <-> :'v777' LIMIT 10;
```
Ngalstyan4, line 19: Can you also add inserts here, to make sure once can continue inserting elements into the table, when the index was first created externally?

test/sql/hnsw_index_from_file.sql
```diff
@@ -0,0 +1,30 @@
...
+DROP INDEX hnsw_l2_index;
+
+-- Validate that creating an index from file works with cosine distance function
+CREATE INDEX hnsw_cos_index ON sift_base1k USING hnsw (v dist_cos_ops) WITH (dims=128, M=16, ef=64, ef_construction=128, _experimental_index_path='/tmp/lanterndb/files/index-sift1k-cos.usearch');
+SELECT * FROM ldb_get_indexes('sift_base1k');
+
+SELECT v AS v777 FROM sift_base1k WHERE id = 777 \gset
+EXPLAIN (COSTS FALSE) SELECT ROUND(cos_dist(v, :'v777')::numeric, 2) FROM sift_base1k order by v <-> :'v777' LIMIT 10;
+SELECT ROUND(cos_dist(v, :'v777')::numeric, 2) FROM sift_base1k order by v <-> :'v777' LIMIT 10;
+
```
Ngalstyan4, line 30: Consider these cases:

case 1:

1) Index is created externally.
2) More vectors are added to the table
3) CREATE INDEX is run on the table with the external file

case 2:
1) Index is created externally
2) Vectors are deleted from the table
3) CREATE iNDEX is run on the table with external file

case 3:
1) Index is created externally
2) Vectors updated
3) CREATE iNDEX is run on the table with external file


I assume currently case1 results in certain vectors not being in the index and never being returned by search queries. If so, let's add the test to the `todos` regression test so we have it tracked. If the case is handled, let's have a test for it.

I think case2 is currently handled fine. Our index will have extra dead entries but that is fine. We can deal with it when we implement proper FULL VACUUM. Can you check this and add a test case for it?

case 3: I think we should explicitly ban this in external index API definition in our docs. That said, I think this works fine in most cases since I think UPDATEs result in a new tuple being created with a fresh ctid. (Things probably break after a VACUUM that is why we should ban this in docs)

src/hnsw/build.c
var77: Yes I will do the required changes on a separate PR and remove this comments once done.

src/hnsw/options.c
var77: I didn't as I don't have windows machine, but I found it on stackoverflow that the `unistd.h` does not exist in windows. https://stackoverflow.com/questions/37437082/which-is-a-good-practice-to-use-access-function

test/sql/hnsw_index_from_file.sql
var77: Yes, currently this params are required for index scan to work correctly.

src/hnsw/options.c
```diff
@@ -54,6 +60,17 @@ int HnswGetEf(Relation index)
...
 
+char *HnswGetIndexFilePath(Relation index)
+{
+    HnswOptions *opts = (HnswOptions *)index->rd_options;
+    if(!opts) return NULL;
+    if(!opts->experimantal_index_path_offset) {
+        return NULL;
+    }
+
+    return (char *)opts + opts->experimantal_index_path_offset;
```
Ngalstyan4, line 71: Oh, this makes so much sense - variable length options are stored in the struct indirectly, via an offset.
This is what I had missed when I could not get string options working

test/sql/hnsw_index_from_file.sql
```diff
@@ -0,0 +1,81 @@
...
+SELECT * FROM ldb_get_indexes('sift_base1k');
+
+SET enable_seqscan = false;
+
+SELECT v AS v777 FROM sift_base1k WHERE id = 777 \gset
+EXPLAIN (COSTS FALSE) SELECT ROUND(l2sq_dist(v, :'v777')::numeric, 2) FROM sift_base1k order by v <-> :'v777' LIMIT 10;
+SELECT ROUND(l2sq_dist(v, :'v777')::numeric, 2) FROM sift_base1k order by v <-> :'v777' LIMIT 10;
+-- Validate that inserting rows on index created from file works as expected
+INSERT INTO sift_base1k (v) VALUES 
+('{1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1}'),
```
Ngalstyan4, line 26: this can be `INSERT INTO sift_base1k (id, v) VALUES (1001, '{...}') so it is more clear that the id of the vector is 1001.

And, instead of typing out the vector explicitly, you can probably use another vector from the existing table.
E.g. SELECT v AS v777 FROM sift_base1k WHERE id = 777 \gset
then insert the value :'v777'

test/sql/hnsw_index_from_file.sql
```diff
@@ -0,0 +1,81 @@
...
+-- More vectors are added to the table
+-- CREATE INDEX is run on the table with the external file
+
+DROP TABLE sift_base1k CASCADE;
+\ir utils/sift1k_array.sql
+INSERT INTO sift_base1k (v) VALUES 
+('{1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1}'), 
+('{2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2}');
+CREATE INDEX hnsw_l2_index ON sift_base1k USING hnsw (v dist_l2sq_ops) WITH (dims=128, M=16, ef=64, ef_construction=128, _experimental_index_path='/tmp/lanterndb/files/index-sift1k-l2.usearch');
+-- The 1001 and 1002 vectors will be ignored in search, so the first row will not be 0 in result
```
Ngalstyan4, line 56: This test should be in `hnsw_todos`

test/sql/hnsw_index_from_file.sql
```diff
@@ -0,0 +1,81 @@
...
+
+-- Case 3:
+-- Index is created externally
+-- Vectors updated
+-- CREATE INDEX is run on the table with external file
+DROP TABLE sift_base1k CASCADE;
+\ir utils/sift1k_array.sql
+UPDATE  sift_base1k SET v=:'v1001' WHERE id=777;
+CREATE INDEX hnsw_l2_index ON sift_base1k USING hnsw (v dist_l2sq_ops) WITH (dims=128, M=16, ef=64, ef_construction=128, _experimental_index_path='/tmp/lanterndb/files/index-sift1k-l2.usearch');
+-- The first row will not be 0 now as the vector under id=777 was updated to 1,1,1,1... but it was indexed with different vector
```
Ngalstyan4, line 78: this test should be in `hnsw_todos`

Ngalstyan4 (approved):
Code looks good. Left comments on tests

test/sql/hnsw_index_from_file.sql
```diff
@@ -0,0 +1,81 @@
...
+SELECT ROUND(l2sq_dist(v, :'v777')::numeric, 2) FROM sift_base1k order by v <-> :'v777' LIMIT 10;
+-- Validate that inserting rows on index created from file works as expected
+INSERT INTO sift_base1k (v) VALUES 
+('{1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1}'), 
+('{2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2}');
+SELECT v AS v1001 FROM sift_base1k WHERE id = 1001 \gset
+SELECT ROUND(l2sq_dist(v, :'v1001')::numeric, 2) FROM sift_base1k order by v <-> :'v1001' LIMIT 10;
+
+-- Drop and recreate table
+DROP TABLE sift_base1k CASCADE;
```
dqii, line 32: Not a big deal, but is it necessary to drop the table? You could just drop the index.

test/sql/hnsw_index_from_file.sql
dqii: Caveat: I understand that case 3 should be banned for now until deletes and inserts are working.

I would think that case 3 works as long as case 1 and 2 work, since update = delete and insert as far as Postgres is considered. The index would have a pointer to a dead tuple, which would be vacuumed. Whatever means we use to add newly inserted vectors in case 1, I would expect to also apply to case 3.

test/sql/hnsw_index_from_file.sql
var77: as I have added some rows to the table, I was thinking it would be clear to drop the table, so the new test case won't be messed with this one.

dqii (approved):

PR #100, opened 2023-08-31
Add benchmarking to local dev and to CI/CD

dqii:
image


dqii:
Just documenting the current state.

`make benchmark` runs locally for me.

The `pip install`'s don't work in CI/CD unless you remove `--break-system-packages`. This breaks local for me.
If you remove `--break-system-packages`, it still doesn't work because the `DATABASE_URL`'s are wrong.

dqii:
Update:
- `make benchmark` runs locally for me.

Two issues:
- The database URLs are incorrect in `run_benchmarks` and cause CI/CD to not work.
- Despite the fact that CI/CD didn't work, it's marked as passing when it shouldn't be.

dqii:
All issues are fixed.

var77 (approved):

PR #101, opened 2023-09-02
Save and load usearch options from index header page

var77:
## Description
With this changes we will keep the necessary usearch options (`dimensions`, `ef`, `ef_construction`, `m`, `metric_kind`) in our index header when building index and in scans read the options from index header instead of index options.
Also we will add `ef` and `ef_construction` params in usearch index header, and make it reload the values when loaded from file as you can see in [this PR](https://github.com/Ngalstyan4/usearch/pull/8) . 

By doing this we can omit all the parameters when building index from file, as we will get the required values from the usearch header and save them in our index header to be used on table scans.

Currently the tests are failing, as the old versions of index files are being used, they need to be replaced in the bucket with newer versions.

### DIFF
--- a/src/hnsw/build.c
+++ b/src/hnsw/build.c
@@ -330,16 +330,18 @@ static void BuildIndex(
     buildstate->hnsw = NULL;
     if(buildstate->index_file_path) {
         usearch_load(buildstate->usearch_index, buildstate->index_file_path, &error);
-        assert(error == NULL);
+        if(error != NULL) {
+            elog(ERROR, "%s", error);
+        }
         elog(INFO, "done loading usearch index");
 
-        // todo determine how we can set index rd_options
-        // HnswOptions *opts = (HnswOptions *)index->rd_options;
-        // opts->m = usearch_connectivity(buildstate->usearch_index, &error);
-        // assert(error == NULL);
-        // opts->dims = usearch_dimensions(buildstate->usearch_index, &error);
-        // assert(error == NULL);
-        // todo determine how to get ef construction and ef search params
+        usearch_metadata_t metadata = usearch_metadata(buildstate->usearch_index, &error);
+        assert(error == NULL);
+        opts.connectivity = metadata.connectivity;
+        opts.dimensions = metadata.dimensions;
+        opts.expansion_add = metadata.expansion_add;
+        opts.expansion_search = metadata.expansion_search;
+        opts.metric_kind = metadata.metric_kind;
     } else {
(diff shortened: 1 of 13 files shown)


src/hnsw/build.c
```diff
@@ -333,13 +333,16 @@ static void BuildIndex(
...
-        // opts->m = usearch_connectivity(buildstate->usearch_index, &error);
-        // assert(error == NULL);
-        // opts->dims = usearch_dimensions(buildstate->usearch_index, &error);
-        // assert(error == NULL);
-        // todo determine how to get ef construction and ef search params
+        opts.connectivity = usearch_connectivity(buildstate->usearch_index, &error);
+        assert(error == NULL);
+        opts.dimensions = usearch_dimensions(buildstate->usearch_index, &error);
+        assert(error == NULL);
+        opts.expansion_add = usearch_expansion_add(buildstate->usearch_index, &error);
```
Ngalstyan4, line 340: Could all of these be retrieved from `usearch_metadata`?

test/expected/hnsw_index_from_file.out
```diff
@@ -39,7 +39,7 @@ EXPLAIN (COSTS FALSE) SELECT ROUND(l2sq_dist(v, :'v777')::numeric, 2) FROM sift_
 
 SELECT ROUND(l2sq_dist(v, :'v777')::numeric, 2) FROM sift_base1k order by v <-> :'v777' LIMIT 10;
 INFO:  began scanning with 0 keys and 1 orderbys
-INFO:  starting scan with dimensions=128 M=16 efConstruction=128 ef=64
+INFO:  starting scan with dimensions=128 M=16 efConstruction=64 ef=32
```
Ngalstyan4, line 42: Can you comment on what happened here?

Ngalstyan4 (approved):
Looks good!

test/expected/hnsw_index_from_file.out
var77: I have changed the ef search and ef construction to other values from the default ones we have now, so it would it more visible that the values are being loaded from the file header

src/hnsw/build.c
var77: There should be done some changes in usearch to return this data from metadata but yes technically it is possible.

var77: I have added this params to `usearch_metadata` in [this](https://github.com/Ngalstyan4/usearch/pull/11) PR

PR #103, opened 2023-09-03
Incorporate HNSW-specific information to cost estimate

### LINKED ISSUE #14: Improve Hnsw cost estimate
Properly implement postgres index [cost estimation](https://github.com/lanterndata/lanterndb/blob/65353dc78908f19d63de4e8b050c2b59860d7fca/src/hnsw.c#L44-L126) for postgres query planner

Resources

- Index cost estimation`  https://www.postgresql.org/docs/current/index-cost-estimation.html

yoloVoe:
Closes #14.

We do so by:
1. Computing the number of tuples we expect to access on a given search query based on some bounds from the paper.
2. Compute the number of datablock, header and blockmap accesses. Number of header accesses is assumed to be 1 as we expect it to stay in cache as everytime we access a blockmap, we also access header so it should stay in cache. For datablock and header, we follow basic logic from `genericcostestimate` [here](https://github.com/postgres/postgres/blob/d0ec2ddbe088f6da35444fad688a62eae4fbd840/src/backend/utils/adt/selfuncs.c#L6516-L6529), but with some hnsw specific enhancements.
3. Totalcost is the result from `genericcostestimate`, which should incorporate the number of tuples we fed it. We also scale the result based on our updated number of page accesses.

Commit message has more details too.


src/hnsw.c
```diff
@@ -84,10 +176,20 @@ static void hnswcostestimate(PlannerInfo *root,
...
-    *indexSelectivity = costs.indexSelectivity;
+    *indexTotalCost = costs.numIndexPages ? costs.indexTotalCost * (num_blocks_accessed / costs.numIndexPages) : 0;
+    *indexSelectivity = num_tuples_in_index ? costs.numIndexTuples / num_tuples_in_index : 0;
+    // since we try to insert index tuples at the last datablock,
+    // there is no "order" at all that can be assumed.
     *indexCorrelation = 0;
-    *indexPages = costs.numIndexPages;
+    *indexPages = num_blocks_accessed;
+
+    elog(INFO, "Query cost estimator");
```
Ngalstyan4, line 186: Can you change these to `elog(DEBUG5...` and then do something like [this](https://github.com/lanterndata/lanterndb/blob/5c794a424f72260e1ee87e4dfcb0b49424514b57/test/sql/hnsw_select.sql#L20) in the cost estimate tests in which you would like these logs printed?

Ngalstyan4 (commented):
Looks good!

src/hnsw.c
yoloVoe: done

PR #106, opened 2023-09-03
cleanup: get rid of unnecessary markbufferdirty and field

yoloVoe:
This commit does 3 things:
- Get rid of index_rel from ExtraDirtiedBufs as it's not used
- Rename ExtraDirtiedBufs fields to be lower case instead of upper case. Based on Narek, they were upper case before to indicate they could only be touched/modified once across all threads (i.e. the variables were global)
- Get rid of unnecessary MarkBufferDirty. In cases where a buffer is tracked by xlog, GenericXLogFinish will call MarkBufferDirty so we don't need to call it ourselves.  [Source](https://github.com/postgres/postgres/blob/master/src/backend/access/transam/generic_xlog.c#L413)

### DIFF
--- a/src/hnsw/external_index.c
+++ b/src/hnsw/external_index.c
@@ -85,7 +85,7 @@ int CreateBlockMapGroup(
         special->lastId = first_node_index + (blockmap_id + 1) * HNSW_BLOCKMAP_BLOCKS_PER_PAGE - 1;
         special->nextblockno = BufferGetBlockNumber(buf) + 1;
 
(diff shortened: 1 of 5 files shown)


var77:
> Looking at test failure.

Hi @yoloVoe , the test failure was caused as we have replaced the usearch index files with newer versions under the same url, which were being used in an existing tests with old usearch version. This should be fixed now, sorry for inconvenience.

yoloVoe:
Ah, I see. Thanks for the update!

PR #107, opened 2023-09-03
revert last couple of cimmits

Ngalstyan4:


Ngalstyan4:
Thought, these caused a failure, but the root cause was something else (see https://github.com/lanterndata/lanterndb/pull/106#issuecomment-1704406412)

PR #120, opened 2023-09-06
Reduce elog overhead on fastpath

Ngalstyan4:
Should be merged now.
- [ ] ~Add a CmakeLists option for defining `LANTERNDB_DEBUG_LOGS`~
- [ ] ~Build LanternDB with `LANTERNDB_DEBUG_LOGS` in ci/cd for tests~
- [ ] ~Separate out tests that need `LANTERNDB_DEBUG_LOGS` into one file (This will make sure that if users run tests locally, without enabling this flag on their release branch, tests only in one file will fail)~


Ngalstyan4:
Actually, seems kind of annoying to have yet another build option for this..
Opened an issue (#133) to deal with this in a cleaner way

PR #123, opened 2023-09-07
Enable codecov check only for prs

var77:
This PR will change the codecov default behaviour to run only on pull requests, so we won't have the failure checkmark on main branch in case of codecov failure


var77:
No it will upload the coverage information anyway from our CI, so the badge will be updated. Just the codecov bot won't trigger it's checks on push actions, but it will run on PRs

PR #124, opened 2023-09-08
Throw error when index operator is used without index

dqii:
This implementation uses the ExecutorStart_hook to detect that the usage of the operator <-> is only in the context of an index scan.


dqii:
I considered passing data from the post_parse_analyze_hook to the ExecutorStart_hook indicating that the operator is used, so that we didn't do this check in ExecutorStart_hook unnecessarily. The approaches didn't seem clean. I didn't want global variables. I didn't want a hash table. There's no built in way. I think it's better to just accept the duplication.

I considered removing the parse analyzer. This doesn't work because the planner is missing some information that the parser has.

I considered reducing the number of tree iterations by the post-parse hook by merging the operator check and the sort function. I didn't think this was worth it given that most queries probably won't use the index.

Lastly, more could be done to flesh out the plan tree walker. I opted against it for this diff to keep it simpler and because the other cases were lower priority.

src/hooks/executor_start.c
```diff
@@ -0,0 +1,76 @@
...
+    if(IsA(node, IndexScan)) {
+        context->isIndexScan = true;
+        bool status = plan_tree_walker(node, operator_used_correctly_walker, (void *)context);
+        context->isIndexScan = false;
+        return status;
+    }
+    if(IsA(node, OpExpr)) {
+        OpExpr *opExpr = (OpExpr *)node;
+        if(list_member_oid(context->oidList, opExpr->opno) && !context->isIndexScan) {
+            return true;
```
Ngalstyan4, line 30: Does this function return 'true' when the operator is used correctly or when the operator is used incorrectly?

src/hooks/executor_start.c
```diff
@@ -0,0 +1,76 @@
...
+{
+    List *oidList;
+    bool  isIndexScan;
+} OperatorUsedCorrectlyContext;
+
+static bool operator_used_correctly_walker(Node *node, OperatorUsedCorrectlyContext *context)
+{
+    if(node == NULL) return false;
+    if(IsA(node, IndexScan)) {
+        context->isIndexScan = true;
```
Ngalstyan4, line 22: Can you explain what's happening here?
why do we set 'isIndexScan' to true, and then back to false on line 24

src/hooks/executor_start.c
```diff
@@ -0,0 +1,76 @@
...
+{
+    OperatorUsedCorrectlyContext context;
+    context.oidList = oidList;
+    context.isIndexScan = false;
+    if(operator_used_correctly_walker(plan, &context)) {
+        elog(ERROR, "Operator <-> has no standalone meaning and is reserved for use in vector index lookups only");
+    }
+}
+
+ExecutorStart_hook_type original_ExecutorStart_hook = NULL;
```
Ngalstyan4, line 59: Let's deine the global variable in the beginning of the file, before any function definitions

src/hooks/utils.c
```diff
@@ -0,0 +1,29 @@
+#include <postgres.h>
+
+#include <catalog/pg_type_d.h>
+#include <nodes/makefuncs.h>
+#include <nodes/pg_list.h>
+
+List *get_operator_oids()
```
Ngalstyan4, line 7: Can you rename this or prefix with `ldb_` 
At function callsite, I thought this is a postgres native function, and was confused why we wanted all operator oids.

src/hooks/post_parse.c
```diff
@@ -0,0 +1,146 @@
...
+{
+    if(original_post_parse_analyze_hook) {
+#if PG_VERSION_NUM >= 140000
+        original_post_parse_analyze_hook(pstate, query, jstate);
+#else
+        original_post_parse_analyze_hook(pstate, query);
+#endif
+    }
+
+    List *oidList = get_operator_oids();
```
Ngalstyan4, line 137: oidList can be NULL here right?
If we account for that in function calls below, maybe leave a comment here to make it clear we did not forget it

src/hooks/post_parse.c
```diff
@@ -0,0 +1,146 @@
...
+
+    return expression_tree_walker(node, operator_used_correctly_walker, (void *)context);
+}
+
+static bool is_operator_used_correctly(Node *node, List *oidList, List *sortGroupRefs)
+{
+    OperatorUsedCorrectlyContext context;
+    context.oidList = oidList;
+    context.sortGroupRefs = sortGroupRefs;
+    return !operator_used_correctly_walker(node, &context);
```
Ngalstyan4, line 117: This function has very strange semantics:

'is_operator_used_correctly' returns the opposite of 'operator_used_correctly_walker' ?

test/sql/hnsw_select.sql
```diff
@@ -9,8 +9,18 @@ CREATE INDEX ON small_world USING hnsw (v) WITH (dims=3, M=5, ef=20, ef_construc
...
 
+CREATE TABLE test1 (id SERIAL, v REAL[]);
+CREATE TABLE test2 (id SERIAL, v REAL[]);
+INSERT INTO test1 (v) VALUES ('{5,3}');
+INSERT INTO test2 (v) VALUES ('{5,4}');
+CREATE INDEX ON test1 USING hnsw (v);
+
 SET enable_seqscan = false;
 
+-- Verify that basic queries still work
```
Ngalstyan4, line 20: add to comment: 
-- Verify that basic queries still work given our query parser and planner hooks

test/sql/utils/common.sql
```diff
@@ -41,3 +41,20 @@ BEGIN
...
 LANGUAGE plpgsql;
+
+-- Determines if the provided SQL query uses an "Index Scan" by examining its EXPLAIN plan.
+-- Reason: The output of EXPLAIN is not consistent across postgres versions
+CREATE OR REPLACE FUNCTION has_index_scan(query_text text) RETURNS boolean AS $$
+DECLARE
+    plan_row RECORD;
+    found boolean := false;
+BEGIN
+    FOR plan_row IN EXECUTE 'EXPLAIN ' || query_text LOOP
```
Ngalstyan4, line 52: I think the function would be clearer if EXPLAIN was passed with the query at callsite.
Then, it would be clear this function just inspects the output of the EXPLAIN query.

test/sql/hnsw_select.sql
```diff
@@ -36,11 +46,28 @@ WITH neighbors AS (
 RESET client_min_messages;
 
 -- Verify where condition works properly and still uses index
-SELECT * FROM small_world WHERE b IS TRUE ORDER BY v <-> '{0,0,0}';
-EXPLAIN (COSTS FALSE) SELECT * FROM small_world WHERE b IS TRUE ORDER BY v <-> '{0,0,0}';
+SELECT has_index_scan('SELECT * FROM small_world WHERE b IS TRUE ORDER BY v <-> ''{0,0,0}''');
```
Ngalstyan4, line 49: Are multiple single quotes in `''{0,0,0}''` intentional?

src/hooks/utils.c
```diff
@@ -0,0 +1,29 @@
...
+{
+    List *oidList = NIL;
+
+    Oid intArrayOid = INT4ARRAYOID;
```
Ngalstyan4, line 11: I'd use these constants directly in the function call below, no need for the local variables. but this is fine too

src/hooks/plan_tree_walker.c
```diff
@@ -0,0 +1,106 @@
...
+    if(walker_func(plan->qual, context)) return true;
+    if(walker_func(plan->lefttree, context)) return true;
+    if(walker_func(plan->righttree, context)) return true;
+    if(walker_func(plan->initPlan, context)) return true;
+    return false;
+}
+
+bool plan_tree_walker(Plan *plan, bool (*walker_func)(Plan *plan, void *context), void *context)
+{
+    check_stack_depth();
```
Ngalstyan4, line 20: what does this do?

src/hooks/plan_tree_walker.c
```diff
@@ -0,0 +1,106 @@
+#include <postgres.h>
+
+#include <miscadmin.h>
+#include <nodes/nodeFuncs.h>
+#include <nodes/parsenodes.h>
+#include <nodes/plannodes.h>
+
+bool plan_tree_walker_util(Plan *plan, bool (*walker_func)(Plan *plan, void *context), void *context)
```
Ngalstyan4, line 8: This function is like `any` over a collection type that checks the predicate (walker_func) in this case
I'd name this
`bool plan_tree_walker_any(Plan *plan, bool (*predicate_func)(Plan *plan, void *context), void *context)`

other names work too, but would be good if the name communicates when the function returns true


`

Ngalstyan4 (commented):
Looks good!

src/hooks/executor_start.c
dqii: Used incorrectly

src/hooks/executor_start.c
dqii: It's traversing a tree. When it hits an IndexScan node, it's going to "go inside it". I'm noting that we're inside an index scan now. If we hit the Lantern operator inside an index scan, that's an acceptable way to use the operator. We return false. If we hit the Lantern operator and we are not inside an index scan, that is not an acceptable way to use the operator. The `if(list_member_oid(context->oidList, opExpr->opno) && !context->isIndexScan) {` will catch this, and return true.

test/sql/hnsw_select.sql
dqii: Yes. Otherwise, it would match with the single quote that starts the string.

src/hooks/post_parse.c
dqii: Walkers return false in general. They return true when you exit prematurely. In my case, I want to exit prematurely when something is wrong, i.e., the operator is used incorrectly.

I could rename everything to "incorrectly"

src/hooks/utils.c
dqii: Yes, done.

src/hooks/executor_start.c
dqii: Yes, done.

src/hooks/utils.c
dqii: Done. This was a relic of an older version of the code.

src/hooks/plan_tree_walker.c
dqii: This is in Postgres's expression_tree_walker function. It checks that we aren't going into too deep a recursion. So would be triggered for an unreasonable complex query, or if there's somehow an infinite loop.

src/hooks/post_parse.c
dqii: Renamed.

src/hooks/executor_start.c
dqii: Renamed for clarity.

src/hooks/plan_tree_walker.c
dqii: `plan_tree_walker` is for plan nodes. `plan_tree_walker_util` is for `Plan`, which is a specific type of Plan node that is shared across all plan nodes. The naming isn't the best, I agree. Will adjust.

test/sql/hnsw_select.sql
dqii: Agree, done.

src/hooks/post_parse.c
dqii: Discussed offline. Will add ldb assert after that diff is pushed.

PR #130, opened 2023-09-10
Cleanup and asserts

Ngalstyan4:


var77:
By this change, should we use `ldb_invariant` instead of `elog(ERROR)` in all places?

Ngalstyan4:
No, some errors are expected and we can actually trigger them in tests. 
Those should continue using `elog` directly

`ldb_invariant` is for invariants that hold no matter what and we can think of no way to trigger these.
The only reason I still `elog` from `ldb_invariant` is to get some hints on what went wrong, in case we do run into a broken invariant.

dqii:
TODO: Add an invariant check to `ldb_ get_operator_oids`

var77 (approved):

src/hooks/plan_tree_walker.c
```diff
@@ -55,7 +57,7 @@ bool plan_tree_walker(Plan *plan, bool (*walker_func)(Node *plan, void *context)
         // Join nodes
         case T_Join:
             Join *join = (Join *)plan;
-            if(base_plan_walker(&(join->plan), walker_func, context)) return true;
+            if(base_plan_walker((Plan *)&(join->plan), walker_func, context)) return true;
```
dqii, line 60: I don't think this or the one below should be necessary, join->plan is always a plan.

dqii (approved):

src/hooks/plan_tree_walker.c
Ngalstyan4: You are right. removed it.

PR #137, opened 2023-09-13
Rename lanterndb->lantern

Ngalstyan4:


var77:
Maybe we can remove the update files, they're not going to work right?

var77 (approved):

Ngalstyan4:
Removing the latest one in #135 
Keeping the rest for reference for when we add update tests to ci/cd.

PR #143, opened 2023-09-15
Add Mac builds into CI pipeline

var77:
## Description
- Added macos-13 runner which will build macos binaries. The build script was separated into 3 parts:
`build.sh` - this is common script for both linux and mocos platforms. This will be executed from the pipeline
`build-linux.sh` - linux specific environment setup functions
`build-mac.sh` - macos specific environment setup functions

now the universal archive will have the following structure `$arch/$platform/$pg_version/lantern.so`

The command to get architecture changed from `dpkg --print-architecture` to `uname -m`, so we will have the same output for both linux and macos. Previously the `dpkg` was printing `amd64` and `uname -m` `x86_64`.

Cleaned up the build scripts and action yaml files.

I will also open PR on `lantern_extras` to add macos support

#7

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -16,51 +16,40 @@ on:
         required: false
         default: false
 jobs:
-  ubuntu-build:
+  build:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - postgres: 16
-            os: ubuntu-22.04
-          - postgres: 15
-            os: ubuntu-22.04
-          - postgres: 14
-            os: ubuntu-22.04
-          - postgres: 13
-            os: ubuntu-22.04
-          - postgres: 12
-            os: ubuntu-22.04
-          - postgres: 11
-            os: ubuntu-22.04
+        os: [ubuntu-22.04, macos-13]
+        postgres: [11, 12, 13, 14, 15, 16]
     steps:
       - uses: actions/checkout@v3
         with:
           submodules: "recursive"
       - name: Build
         id: build
-        run: sudo su -c "PG_VERSION=$PG_VERSION USE_SOURCE=1 GITHUB_OUTPUT=$GITHUB_OUTPUT BUILD_PACKAGES=1 ./ci/scripts/build-linux.sh"
+        run: sudo sh -c "PG_VERSION=$PG_VERSION USE_SOURCE=1 GITHUB_OUTPUT=$GITHUB_OUTPUT BUILD_PACKAGES=1 ./ci/scripts/build.sh"
         env:
           PG_VERSION: ${{ matrix.postgres }}
-          BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
       - name: Create Archive Package
         id: archive
-        run: sudo su -c "GITHUB_OUTPUT=$GITHUB_OUTPUT ./ci/scripts/package-archive.sh"
-      - name: Upload artifacts
+        run: sudo sh -c "GITHUB_OUTPUT=$GITHUB_OUTPUT ./ci/scripts/package-archive.sh"
+      - name: Upload deb package artifacts
         uses: actions/upload-artifact@v3
+        if: ${{ steps.build.outputs.deb_package_path != '' }}
         with:
-          name: ${{ steps.build.outputs.package_name }}
-          path: ${{ steps.build.outputs.package_path }}
+          name: ${{ steps.build.outputs.deb_package_name }}
+          path: ${{ steps.build.outputs.deb_package_path }}
       - name: Upload archive package artifacts
         uses: actions/upload-artifact@v3
-        if: ${{ steps.archive.outputs.archive_package_name != '' }}
+        if: ${{ steps.archive.outputs.archive_package_path != '' }}
         with:
           name: lantern-package
           path: ${{ steps.archive.outputs.archive_package_path }}
-  ubuntu-package:
+  package:
     runs-on: ubuntu-22.04
-    needs: [ubuntu-build]
+    needs: [build]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/download-artifact@v3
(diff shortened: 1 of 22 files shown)


.github/workflows/build.yaml
```diff
@@ -16,49 +16,40 @@ on:
...
+        os: [ubuntu-22.04, macos-13]
+        postgres: [11, 12, 13, 14, 15]
     steps:
       - uses: actions/checkout@v3
         with:
           submodules: "recursive"
       - name: Build
         id: build
-        run: sudo su -c "PG_VERSION=$PG_VERSION USE_SOURCE=1 GITHUB_OUTPUT=$GITHUB_OUTPUT BUILD_PACKAGES=1 ./ci/scripts/build-linux.sh"
+        run: sudo sh -c "PG_VERSION=$PG_VERSION USE_SOURCE=1 GITHUB_OUTPUT=$GITHUB_OUTPUT BUILD_PACKAGES=1 ./ci/scripts/build.sh"
```
dqii, line 32: Was this change deliberate? Was `su` from before a typo?

.github/workflows/test-linux.yaml
```diff
@@ -21,24 +21,15 @@ jobs:
...
+        os: [ubuntu-22.04]
+        postgres: [11, 12, 13, 14, 15]
     steps:
       - uses: actions/checkout@v3
         with:
           submodules: "recursive"
       - name: Build
         id: build
-        run: sudo su -c "PG_VERSION=$PG_VERSION USE_SOURCE=1 GITHUB_OUTPUT=$GITHUB_OUTPUT ENABLE_COVERAGE=1 ./ci/scripts/build-linux.sh"
+        run: sudo su -c "PG_VERSION=$PG_VERSION USE_SOURCE=1 GITHUB_OUTPUT=$GITHUB_OUTPUT ENABLE_COVERAGE=1 ./ci/scripts/build.sh"
```
dqii, line 32: This says su, which is intended?

.github/workflows/build.yaml
var77: `-c` option of `su` is not available on Mac OS, so on the steps which will run on Ma OS I replaced it with `sh -c`

dqii (approved):

Ngalstyan4:
Can you install pgvector in mac as well so our pgvector tests also run?

I think right now those tests will fail because of an innocuous reordering issue in the test.
Feel free to ignore that failure for now

var77:
> Can you install pgvector in mac as well so our pgvector tests also run?
> 
> I think right now those tests will fail because of an innocuous reordering issue in the test. Feel free to ignore that failure for now

@Ngalstyan4  currently I do not run tests on mac as in CI it is collecting coverage information now and the postgres should be run in a custom way with gcov enabled. I should separate that part as well, so the tests will run on mac without coverage information. Should I implement it in this PR?

ci/scripts/build.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function build_and_install() {
+  cd /tmp/lantern
+  mkdir build
+  cd build
+
+  flags="-DUSEARCH_NO_MARCH_NATIVE=ON"
+  
+  # On macs -static-libstdc++ throws warning
```
Ngalstyan4, line 45: Let's actually modify the cmake file to not pass this flag in mac build.
Then we will have one less warning and will get rid of the `if` below

ci/scripts/run-benchmarks.sh
```diff
@@ -20,7 +20,7 @@ export PG_VERSION=15
 export GITHUB_OUTPUT=/dev/null
 export PGDATA=/etc/postgresql/$PG_VERSION/main/
 
-echo "port = 5432" >> $PGDATA/postgresql.conf
+echo "port = 5432" >> ${PGDATA}postgresql.conf
```
Ngalstyan4, line 23: It seems `PGDATA` has a trailing slash, but other path variables do not (e.g. WORKDIR does not)
We should choose and document a convention and make sure it is followed everywhere.

`pg_config` seems to return paths without the traling slash. Example:
```
INCLUDEDIR-SERVER = /usr/include/postgresql/13/server
LIBDIR = /usr/lib/x86_64-linux-gnu
PKGLIBDIR = /usr/lib/postgresql/13/lib
```
So, I am inclined to use that convention here.

ci/scripts/run-benchmarks.sh
```diff
@@ -20,7 +20,7 @@ export PG_VERSION=15
...
 # Enable auth without password
 echo "local   all             all                                     trust" >  $PGDATA/pg_hba.conf
 echo "host    all             all             127.0.0.1/32            trust" >>  $PGDATA/pg_hba.conf
```
Ngalstyan4, line 26: note that if we keep the current value of `PGDATA` , these two lines need to be updated to remove the slash after  `$PGDATA`

scripts/get_arch_and_platform.sh
```diff
@@ -0,0 +1,26 @@
...
+  echo "Could not detect system architecture. Please specify with ARCH env variable"
+  exit 1
+fi
+
+unameOut="$(uname -s)"
+case "${unameOut}" in
+    Linux*)     PLATFORM=linux;;
+    Darwin*)    PLATFORM=mac;;
+    CYGWIN*)    PLATFORM=cygwin;;
+    MINGW*)     PLATFORM=mingw;;
```
Ngalstyan4, line 20: when do we expect `cygwin`, `mingw` cases to be hit?

Ngalstyan4 (approved):
We can merge this and then do testing on mac in a separate PR

scripts/get_arch_and_platform.sh
var77: they seems to be some unix like shells for windows, but I didn't test them.

var77:
@Ngalstyan4 @dqii I did the requested changes and also added tests for mac os (with pgvector). There needed to do some modifications on pgvector test file to make it more consistent over the platforms.

dqii:
It still lgtm. Sorry, I merged Postgres 16 so there are a few conflicts

var77:
> It still lgtm. Sorry, I merged Postgres 16 so there are a few conflicts

No worries, I will fix them later today

PR #155, opened 2023-09-18
Check for undefined symbols #145

### LINKED ISSUE #145: Check for undefined symbols in our shared library that will not be provided by postgres in CI/CD
context:
detailed description of the issue in the comment here:
https://github.com/lanterndata/lantern/blob/main/scripts/extern_defined.sh#L5-L34
https://github.com/lanterndata/lantern/blob/main/scripts/check_symbols.sh

var77:
I have modified `scripts/extern_defined.sh` script to get the postgres binary dependencies dynamic and also get postgres binary path from `pgconfig`.  
Now there will be check in CI for undefined symbols in our binary.

For test I have added undefined function in our hnsw.h header file and the called that function from hnsw.c file. The check is returning the missing symbol correctly
Pasted Graphic

closes #145

### DIFF
--- a/ci/scripts/build-linux.sh
+++ b/ci/scripts/build-linux.sh
@@ -37,7 +37,12 @@ function package_if_necessary() {
 }
 
 function cleanup_environment() {
+  # Check for undefined symbols
+  if [ ! -n "$ENABLE_COVERAGE" ]
+  then
+    /tmp/lantern/scripts/check_symbols.sh ./lantern.so
+  fi
+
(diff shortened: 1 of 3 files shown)


Ngalstyan4:
Please comment each line with some info on why we need it. the old code comments could help.

Also, please test locally that if you create an undefined function, this check fails, even when compilation of the shared library succeeds.

Otherwise, looks good!

Ngalstyan4 (approved):
Looks good!

Issue #92, opened 2023-08-27
vector_l2sq_dist doesn't seem necessary


dqii:
l2sq_dist works on pgvector's vector type.
e.g., `SELECT l2sq_dist('[1,1]'::vector, '[0,0]'::vector);`
and `SELECT vector_l2sq_dist('[1,1]'::vector, '[0,1]'::vector);`

PR #157, opened 2023-09-18
remove pgvector specific l2sq metric

ezra-varady:
This removes the extra function to calculate L2 squared distance specifically on pgvectors vector type, addressing #92. It also cleans up what I believe is a superflous condition in the sql file. In case it's useful, [this](https://github.com/pgvector/pgvector/blob/b247b688a8ef56103b2dea1f9dd53a87b27696b6/sql/vector.sql#L145-L146) is where pgvector defines the implicit cast of their datatype to real

### DIFF
--- a/ci/scripts/run-benchmarks.sh
+++ b/ci/scripts/run-benchmarks.sh
@@ -30,7 +30,7 @@ POSTGRES_HOST_AUTH_METHOD=trust /usr/lib/postgresql/$PG_VERSION/bin/postgres 1>/
 wait_for_pg
 cd $WORKDIR/build
 
-export DATABASE_URL=postgresql://localhost:5432/postgres
+export LANTERN_DATABASE_URL=postgresql://localhost:5432/postgres
 git clone https://github.com/lanterndata/benchmark
 cd benchmark
 pip install -r core/requirements.txt
--- a/ci/scripts/run-benchmarks.sh
+++ b/ci/scripts/run-benchmarks.sh
@@ -15,7 +15,7 @@ wait_for_pg(){
  done
 }
 
-export WORKDIR=/tmp/lantern
+export WORKDIR=/home/vagrant/lanterndb
 export PG_VERSION=15
 export GITHUB_OUTPUT=/dev/null
 export PGDATA=/etc/postgresql/$PG_VERSION/main/
@@ -31,6 +31,7 @@ wait_for_pg
 cd $WORKDIR/build
 
 export LANTERN_DATABASE_URL=postgresql://localhost:5432/postgres
+export DATABASE_URL=postgresql://localhost:5432/postgres
 git clone https://github.com/lanterndata/benchmark
 cd benchmark
 pip install -r core/requirements.txt
--- a/ci/scripts/run-benchmarks.sh
+++ b/ci/scripts/run-benchmarks.sh
@@ -15,7 +15,7 @@ wait_for_pg(){
  done
 }
 
-export WORKDIR=/home/vagrant/lanterndb
+export WORKDIR=/tmp/lantern
 export PG_VERSION=15
 export GITHUB_OUTPUT=/dev/null
 export PGDATA=/etc/postgresql/$PG_VERSION/main/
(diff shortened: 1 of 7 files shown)


ezra-varady:
not sure what's causing the increase in latency, I don't see it locally

ci/scripts/run-benchmarks.sh
```diff
@@ -30,6 +30,7 @@ POSTGRES_HOST_AUTH_METHOD=trust /usr/lib/postgresql/$PG_VERSION/bin/postgres 1>/
 wait_for_pg
 cd $WORKDIR/build
 
+export LANTERN_DATABASE_URL=postgresql://localhost:5432/postgres
```
dqii, line 33: If you pull from main this should no longer be necessary. Sorry about that

sql/lantern.sql
```diff
@@ -29,14 +29,13 @@ BEGIN
 	END IF;
 
 	IF pgvector_exists THEN
-		-- taken from pgvector so our index can work with pgvector types
-		CREATE FUNCTION vector_l2sq_dist(vector, vector) RETURNS float8
-			AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+                CREATE FUNCTION l2sq_dist(real[], real[]) RETURNS real
```
dqii, line 32: Is it necessary to define this twice? I notice it in line 60 as well.

test/expected/hnsw_vector.out
```diff
@@ -111,12 +111,12 @@ FROM small_world ORDER BY v <-> '[0,1,0]'::VECTOR LIMIT 7;
  100 | 2.00
 (7 rows)
 
-EXPLAIN SELECT id, ROUND(vector_l2sq_dist(v, '[0,1,0]'::VECTOR)::numeric, 2) as dist
+EXPLAIN SELECT id, ROUND(l2sq_dist(v, '[0,1,0]'::VECTOR)::numeric, 2) as dist
```
dqii, line 114: Can you add COSTS FALSE here? So `EXPLAIN (COSTS FALSE) SELECT`

dqii:
LGTM overall

ezra-varady:
@dqii I made the requested changes! I apologize if I'm misunderstanding, but it appears the environment variable is undefined in the upstream ci script

sql/lantern.sql
```diff
@@ -29,14 +45,11 @@ BEGIN
...
-		-- taken from pgvector so our index can work with pgvector types
-		CREATE FUNCTION vector_l2sq_dist(vector, vector) RETURNS float8
-			AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
-
+                -- Define operator class for vector type
 		CREATE OPERATOR CLASS dist_vec_l2sq_ops
 			DEFAULT FOR TYPE vector USING lantern_hnsw AS
 			OPERATOR 1 <-> (vector, vector) FOR ORDER BY float_ops,
-			FUNCTION 1 vector_l2sq_dist(vector, vector);
+			FUNCTION 1 l2sq_dist(real[], real[]);
```
Ngalstyan4, line 52: I think this works because now every time the operator is invoked, the inputs of type `vector` are implicitly cast to type `real[]`. Is this the case?
If so, how expensive is the cast? Can you point to the pgvector code implementing the cast? Can we get away without the cast by just defining a C function that understands the `vector` type and directly extracts floats from there for distance comparison?

context: 
- type casts https://www.postgresql.org/docs/current/sql-createcast.html
- custom types https://www.postgresql.org/docs/current/sql-createtype.html

Ngalstyan4 (commented):
Do you know about postgres update scripts?

Can you create a 0.0.4-latest.sql update script for the changes you have here?

ci/scripts/run-benchmarks.sh
dqii: Can you do one more merge from main? Fixed the upstream CI/CD

PR #158, opened 2023-09-18
Support Postgres 16

dqii:


var77 (approved):

src/hooks/plan_tree_walker.c
```diff
@@ -4,9 +4,19 @@
 
 #include <miscadmin.h>
 #include <nodes/nodeFuncs.h>
+#include <nodes/nodes.h>
 #include <nodes/parsenodes.h>
 #include <nodes/plannodes.h>
 
+bool is_plan_tree_node(Node *node)
```
Ngalstyan4, line 11: I think if you make this `static inline` and put it in the header file (`src/hooks/plan_tree_walker.h`) it will be more likely to be inlined by the complier

src/hooks/plan_tree_walker.c
```diff
@@ -63,15 +72,15 @@ bool plan_tree_walker(Plan *plan, bool (*walker_func)(Node *plan, void *context)
             if(base_plan_walker(&(ctescan->scan.plan), walker_func, context)) return true;
             break;
         }
-        // Join nodes
+#if PG_VERSION_NUM < 160000
         case T_Join:
```
Ngalstyan4, line 76: what hapened to T_Join in postgres 16?

test/expected/hnsw_cost_estimate.out
```diff
@@ -1,116 +1,129 @@
...
+    line text;
+BEGIN
+    EXECUTE explain_query INTO explain_output;
+    FOR line IN (
+        SELECT
+            unnest(string_to_array(explain_output, E'\n')))
+        LOOP
+            IF position(' ' IN
+            LEFT (line, 1)) = 0 AND line LIKE '%cost=%' THEN
+                cost_match :=(regexp_matches(line, 'cost=\d+\.\d+..\d+\.\d+'))[1];
```
Ngalstyan4, line 24: I find it easier to read regex when there is an example string you are trying to match in a comment.
Maybe add an example cost estimate and show how these lines break it down?

Ngalstyan4 (approved):
added small comments. Looks good!
Can you also delete [this folder (test/sql/results)](https://github.com/lanterndata/lantern/tree/main/test/sql/results) from source as part of the PR as well?

I had accidentally committed it at some point.

src/hooks/plan_tree_walker.c
dqii: From what I could tell from playing around with EXPLAINs, it stopped being used by Postgres 15. JOINs were handled with another node (I think T_Group?) in Postgres 15. So I think they just dropped it because not being used.

Issue #133, opened 2023-09-10
Create a general shared variable to communicate DB internal state to SQL for tests


Ngalstyan4:
We want to make sure our regression tests cover different scenarios in which a feature can be used.
These different scenarios do not always have SQL-visible output. For these cases we currently rely on adding `elog(DEBUG5,..)` statements and make sure those `elog`s are only enabled for the test under question. 

See PRs #78 and #103 For more details on how this kind of testing works. In #78, for example, with this kinds of elogs we make sure that for different values of `init_k` `usearch_search` is triggered expected number of times.

This has downsides
- `elog`s have performance impact even if their output is not logged anywhere.
- Postgres itself prints logs at the enabled level which at various points result in sporadic, unrelated, benign regressions

It would be better to add a special variable on an internal schema `(e.g. _lanterndb_internal.state_for_tests) that can be used to comminicate data to our regression tests. In tests where we want to enable this kind of communication we would set the variable and print it from the regression test

PR #159, opened 2023-09-19
initial support for a GUC to control whether logging occurs

ezra-varady:
This adds initial support for a GUC to manage logging, attempting to address #133. It's higher overhead than removing the logging code on release, so it may be undesireable, but it means that we don't need two separate logging macros. If we remove the logging when a release build is done, the release build doesn't pass tests

I understand that while minimal this is a performance regression, so if it's preferable to completely remove unused code from the hot path, it's easy to make a second macro. I chose this because I felt the overhead would be acceptable and it makes the API more consistent

### DIFF
--- a/src/hnsw/utils.h
+++ b/src/hnsw/utils.h
@@ -2,6 +2,7 @@
 #define LDB_HNSW_UTILS_H
 #include <access/amapi.h>
 
+#include "options.h"
 #include "usearch.h"
 
 void            LogUsearchOptions(usearch_init_options_t *opts);
@@ -31,11 +32,11 @@ static inline void ldb_invariant(bool condition, const char *msg, ...)
 // To print debug or test output on these hot codepaths, use ldb_dlog.
 // ldb_dlog has assert() semantics - it will be removed completely on release builds
 // so has no performance impact
-#ifdef LANTERNDB_DEBUG_LOGS
-#define ldb_dlog(elevel, ...) elog(elevel, errmsg_internal(__VA_ARGS__))
-#else
-#define ldb_dlog(elevel, ...) /*empty*/
-
-#endif  // LANTERNDB_DEBUG_LOGS
+#define ldb_dlog(...)                  \
+    {                                  \
+        if(ldb_is_test) {              \
+            elog(DEBUG5, __VA_ARGS__); \
+        }                              \
+    }
 
 #endif  // LDB_HNSW_UTILS_H
--- a/src/hnsw/utils.h
+++ b/src/hnsw/utils.h
@@ -30,11 +30,9 @@ static inline void ldb_invariant(bool condition, const char *msg, ...)
 // search operations.
 // For this reason on hot codepaths we should avoid elog all together.
 // To print debug or test output on these hot codepaths, use ldb_dlog.
-// ldb_dlog has assert() semantics - it will be removed completely on release builds
-// so has no performance impact
 #define ldb_dlog(...)                  \
     {                                  \
-        if(ldb_is_test) {              \
+        if(unlikely(ldb_is_test)) {              \
             elog(DEBUG5, __VA_ARGS__); \
         }                              \
     }
--- a/src/hnsw/utils.h
+++ b/src/hnsw/utils.h
@@ -32,7 +32,7 @@ static inline void ldb_invariant(bool condition, const char *msg, ...)
 // To print debug or test output on these hot codepaths, use ldb_dlog.
 #define ldb_dlog(...)                  \
     {                                  \
-        if(unlikely(ldb_is_test)) {              \
+        if(unlikely(ldb_is_test)) {    \
             elog(DEBUG5, __VA_ARGS__); \
         }                              \
     }
--- a/test/expected/hnsw_cost_estimate.out
+++ b/test/expected/hnsw_cost_estimate.out
@@ -1,18 +1,62 @@
 SET client_min_messages=debug5;
+\ir utils/sift10k_array.sql
+CREATE TABLE IF NOT EXISTS sift_base10k (
+     id SERIAL PRIMARY KEY,
+     v REAL[128]
+);
+\copy sift_base10k (v) FROM '/tmp/lantern/vector_datasets/siftsmall_base_arrays.csv' with csv;
+-- This function executes the given SQL query and returns its estimated total cost.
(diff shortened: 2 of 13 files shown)


src/hnsw/utils.h
```diff
@@ -31,11 +32,11 @@ static inline void ldb_invariant(bool condition, const char *msg, ...)
...
 // so has no performance impact
-#ifdef LANTERNDB_DEBUG_LOGS
-#define ldb_dlog(elevel, ...) elog(elevel, errmsg_internal(__VA_ARGS__))
-#else
-#define ldb_dlog(elevel, ...) /*empty*/
-
-#endif  // LANTERNDB_DEBUG_LOGS
+#define ldb_dlog(...)                  \
+    {                                  \
+        if(ldb_is_test) {              \
```
Ngalstyan4, line 37: I think postgres has `likely()`, `unlikely()` markers to annotate this kind of branches with. probably worth adding an `unlikely` here.

test/expected/hnsw_cost_estimate.out
```diff
@@ -10,6 +10,7 @@ CREATE INDEX hnsw_idx ON sift_base10k_0 USING hnsw (v dist_l2sq_ops) WITH (M=2,
 INFO:  done init usearch index
 INFO:  inserted 0 elements
 INFO:  done saving 0 vectors
+SET _lanterndb_internals.is_test = true;
```
Ngalstyan4, line 13: Let's call the schema `_lantern_internal`

Ngalstyan4 (approved):

ezra-varady:
@Ngalstyan4 made the requested changes and rebased onto main. From what I read there does not seem to be a way to tie a GUC variable's scope to a schema. I'll look over the failing tests in a bit and suppress cost estimates which appear to be the issue

test/sql/hnsw_vector.sql
```diff
@@ -51,14 +51,17 @@ CREATE INDEX ON large_vector USING lantern_hnsw (v);
 \set ON_ERROR_STOP on
 
 -- Validate that index creation works with a larger number of vectors
+SET client_min_messages=debug5;
```
dqii, line 54: I think this file was changed by accident when rebasing. There should be no changes aside from `SET _lantern_internal.is_test = true;` and `SET _lantern_internal.is_test = false;`, and perhaps `SET client_min_messages=debug5;` (it seems like you deliberately added the last one? unsure what it is for)

test/sql/hnsw_cost_estimate.sql
```diff
@@ -1,75 +1,61 @@
...
+     id SERIAL PRIMARY KEY,
+     v real[128]
+);
+\copy sift_base10k_3 (v) FROM '/tmp/lantern/vector_datasets/siftsmall_base_arrays.csv' with csv;
+CREATE INDEX hnsw_idx ON sift_base10k_3 USING hnsw (v dist_l2sq_ops) WITH (M=20, ef_construction=10, ef=16, dim=128);
+SELECT V AS v4444  FROM sift_base10k_3 WHERE id = 4444 \gset
+SET _lantern_internal.is_test = true;
+EXPLAIN (ANALYZE,TIMING FALSE) SELECT * FROM sift_base10k_3 order by v <-> :'v4444'
+LIMIT 10;
+SET _lantern_internal.is_test = false;
```
dqii, line 60: I think this file was also changed by accident when rebasing. There should be no changes other than `SET _lantern_internal.is_test = false;`

ezra-varady: My bad I must have made a mistake while rebasing. fixing it now

var77 (approved):

Ngalstyan4:
Given the new logging approach, I think [this](https://github.com/lanterndata/lantern/blob/main/test/test_runner.sh#L48-L49) is no longer needed. please verify though

ezra-varady:
The only issue encountered when removing that line is that it surfaces several debug logs from within postgres itself. Because these include information about the transaction id they aren't deterministic which breaks the regression tests. In lieu of removing the line that filters for `LANTERN` in debug logs I've modified the `ldb_dlog` macro to automatically prepend it. This makes the slow path somewhat slower but shouldn't affect the fast path

src/hnsw/utils.h
```diff
@@ -29,13 +31,19 @@ static inline void ldb_invariant(bool condition, const char *msg, ...)
...
-#endif  // LANTERNDB_DEBUG_LOGS
+#define ldb_dlog(...)                              \
+    do {                                           \
+        if(unlikely(ldb_is_test)) {                \
+            int   n, size = 200;                   \
+            char *buf = palloc(size);              \
+            n = snprintf(buf, size, __VA_ARGS__);  \
+            if(n >= size) {                        \
+                buf = repalloc(buf, n + 1);        \
+                snprintf(buf, n + 1, __VA_ARGS__); \
```
Ngalstyan4, line 42: Not sure what you are trying to do in the `if` block.

If the passed string interpolation is longer chan 200 chars, you repalloc to 201 chars  and re-interpolate the string.
Are you trying to make sure the string is NULL-terminated here? If so, you should pass `n` at line 42 and set the last char to `NULL`, explicitly. With the current approach, if the string interpolation result is more than 200 chars long, then the result at line 42 will not be null terminated either.

A shorter way to achieve the same would be to pass `size-1` at line 39 and do `palloc0()` at line 38 (`palloc` does not guarantee the returned memory will be zero-initialized)

I think it is much better to skip prepending LANTERN to messages and avoid the whole ordeal altogether.

ezra-varady: The goal is to allow handling debug logs of arbitrary length. `snprintf` should always append a null character I think. The if statement is to allocate extra space if the formatted output would be longer than 199 characters. `n` should hold the number of characters which would be generated for the given input, excluding the trailing null. 

It is definitely an ordeal just to log but I wanted to make it so that manually prepending `LANTERN` to debug logs wasn't required to surface them in regression tests

Ngalstyan4: > n should hold the number of characters which would be generated for the given input

I do not think this is true
see `snprintf` docs here: https://linux.die.net/man/3/snprintf

> **Return value**
> Upon successful return, these functions return the number of characters printed (excluding the null byte used to end output to strings).

Can you get rid of the whole prepending though? not worth it. You can add "LANTERN" to the string at callsite if you'd like

PR #160, opened 2023-09-19
Replace fa_cache with htable cache

var77:
Replaced `fa_cache` with `htable` cache for nodes when doing inserts and scans. 
Renamed `Cache` to `BlockNumberCache` and created `NodeCache` as the entry and key size differ. Maybe there's a way to make some functions generic, you can mention on PR comments
This change should improve performance per benchmarks


src/hnsw/node_cache.h
```diff
@@ -0,0 +1,27 @@
...
+    NodeCacheKey key;
+    void        *value;
+} NodeCacheEntry;
+typedef struct NodeCache
+{
+    HTAB   *htab;
+    HASHCTL hctl;
+} NodeCache;
+
+NodeCache node_cache_create();
```
Ngalstyan4, line 21: Can we refactor these to have one `cache_` api?
It seems the only thing different between the two caches in this PR is the key and values size.
So, we can pass those in create call and have the rest of the arguments be `void*` in the data structure definition.

Maybe we could have a type-specific wrapper around the data structure for each instance (`node_cache` and `bln_cache`) to get some compiler help. Not sure if this is necessary or useful enough, though.

var77: Sure!

var77:
@Ngalstyan4 can you please re-review? I did the requested changes.

src/hnsw/htab_cache.c
```diff
@@ -0,0 +1,53 @@
+#include "htab_cache.h"
+
+HTABCache cache_create(char *name, MemoryContext ctx)
+{
+    HTABCache cache;
+
+    HASHCTL hctl;
+    hctl.keysize = sizeof(HTABCacheKey);
+    hctl.entrysize = sizeof(HTABCacheEntry);
```
Ngalstyan4, line 9: can you add an assertion here that the passed memory context is not the global (I think it is called `TopMemoryContext`). We risk memory leaks if we allocate there.

src/hnsw/htab_cache.c
```diff
@@ -0,0 +1,53 @@
...
+
+    entry = (HTABCacheEntry *)hash_search(cache->htab, key, HASH_ENTER, NULL);
+    entry->value = value;
+}
+
+void cache_destroy(HTABCache *cache)
+{
+    MemoryContext old_context = MemoryContextSwitchTo(cache->hctl.hcxt);
+    hash_destroy(cache->htab);
+    MemoryContextDelete(cache->hctl.hcxt);
```
Ngalstyan4, line 51: Do you delete the memory context you are currently on?
Is that safe?

src/hnsw/htab_cache.h
```diff
@@ -0,0 +1,27 @@
+#ifndef LDB_HNSW_HTAB_CACHE_H
+#define LDB_HNSW_HTAB_CACHE_H
+#include "postgres.h"
```
Ngalstyan4, line 3: fix the braces

src/hnsw/htab_cache.c
```diff
@@ -0,0 +1,53 @@
...
+{
+    HTABCacheEntry *entry;
+
+    entry = (HTABCacheEntry *)hash_search(cache->htab, key, HASH_ENTER, NULL);
+    entry->value = value;
+}
+
+void cache_destroy(HTABCache *cache)
+{
+    MemoryContext old_context = MemoryContextSwitchTo(cache->hctl.hcxt);
```
Ngalstyan4, line 49: Memory context ownership is a little strange here. 
It seems the memory context is constructed externally, by the caller, but it is destroyed internally, by this abstract data type. Is there a reason for this?

src/hnsw/htab_cache.h
```diff
@@ -0,0 +1,27 @@
...
+} HTABCacheEntry;
+
+typedef struct HTABCache
+{
+    HTAB   *htab;
+    HASHCTL hctl;
+} HTABCache;
+
+HTABCache cache_create(char *name, MemoryContext ctx);
+bool      cache_remove(HTABCache *cache, HTABCacheKey *key);
```
Ngalstyan4, line 22: Can you document the assumptions of this ADT in this file?

E.g. currently the caller creates the memory context, passes ownership to this ADT and this ADT destroys it upon `cache_destroy`. 
Also answer questions like: 
Who owns *key and *entry pointers passed into these API functions? Does the ADT make assumptions about how long those pointers live? (if I understand it correctly, the ADT assumes that anything passed in lives longer than the cache ADT itself. This invariant must be followed by the user of the ADT so this interface file is a good place to document it)

src/hnsw/htab_cache.h
```diff
@@ -0,0 +1,27 @@
+#ifndef LDB_HNSW_HTAB_CACHE_H
+#define LDB_HNSW_HTAB_CACHE_H
+#include "postgres.h"
+
+#include <utils/hsearch.h>
+#include <utils/memutils.h>
+
+typedef int HTABCacheKey;
```
Ngalstyan4, line 8: You should use sized integers always (e.g. int32)
Would be great if we could add a ci/cd check for this kinds of things

src/hnsw/htab_cache.c
var77: I was thinking that if I create memory context with the same name they will share the same memory, but I looked through the source code and it seems name is mostly used on errors to display on which memory context the error was appeared. I think it is not critical for us to share the same name of MemoryContext among different cache instances, so I brought back memory context creation inside `cache_create`

src/hnsw/htab_cache.h
var77: Changed to int32 for now, will take a not for CI checks

src/hnsw/htab_cache.h
var77: Added comments [here](https://github.com/lanterndata/lantern/pull/160/files#diff-19338c152ee6b524d410f905b9fe9781d838b4f8b9db6f5d964c609824e31d1eR10-R21)

PR #169, opened 2023-09-20
brew packaging

ezra-varady:
This should update the URL and version in [this](https://github.com/ezra-varady/homebrew-lantern) brew tap. It should also fix an issue where source releases do not include submodules. It will require the addition of a COMMITTER_TOKEN secret under which the commits will be made


var77:
Thank you for the research and provided codes @ezra-varady ! Closing this PR as I did some changes in a relevant one [here](https://github.com/lanterndata/lantern/pull/183)

PR #170, opened 2023-09-20
Changed "expected_number_of_levels" as well as fixed the justification of correctness in the comments.

gnehzza:
Changed "expected_number_of_levels" as well as fixed the justification of correctness in the comments.

Question: Should we type math with latex?

### DIFF
--- a/src/hnsw.c
+++ b/src/hnsw.c
@@ -47,38 +47,38 @@ static char *hnswbuildphasename(int64 phasenum)
 #endif
 
 /*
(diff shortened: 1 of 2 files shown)


ezra-varady:
It looks like this will at most only ever differ by the current implementation by one at the cost of a meaningful increase in complexity. Assuming this derivation is the correct one, we could effectively get the benefits by using `floor(log(x) + .577) + 1`. At this point however we're approximating an approximation. Since this is fairly stale, and the current approximation is 'good enough', differing by only a few percent on large queries, I think we can close this

Issue #112, opened 2023-09-04
Every insert resets prng seed


yoloVoe:
Every insert seems to land at the same level as seed gets reset to default value every time.

Ask is to look into this and fix it.

In particular, make sure to run benchmark code to ensure no regression before and after the fix.

yoloVoe:
Discovered when working on https://github.com/lanterndata/lanterndb/pull/114.

PR #172, opened 2023-09-21
Rewrote usearch's usearch_newnode_level function to use a session-glo…

therealdarkknight:
Fixes [#112](https://github.com/lanterndata/lantern/issues/112).

We rewrite the `usearch_newnode_level` function from `usearch`, which generates a random level for a vector that we wish to insert. 

This not only fixes the seed resetting as described in the issue, but also is an optimization because we use a session-global rng that is seeded by postgres, instead of initializing and seeding our own on every insertion.

**Notes:**
- Technically, usearch_newnode_level returns an int32 but previous code implicitly casted to a uin32. I kept this behavior, so it remains a uint32 here as well.
- The rng is from `common/pg_prng.c` from postgres, as seen [here](https://github.com/postgres/postgres/blob/master/src/common/pg_prng.c). It also has a cross-process random state that is available called `pg_global_prng_state` that we use here.
- Note that technically, this global state needs to be initialized at least once. But, it should always be initialized by this point. In particular, by postmaster within postgres, as seen [here](https://github.com/postgres/postgres/blob/e5975c2daad0327b9faa6b62bcfbf173680d14b9/src/backend/postmaster/postmaster.c#L2634).

**Further todos/improvements:**
- The quantity `inverse_log_connectivity` is constant and can be precomputed ahead of time, instead of computing it every time we insert.

### DIFF
--- a/src/hnsw/insert.c
+++ b/src/hnsw/insert.c
@@ -4,7 +4,9 @@
 
 #include <access/generic_xlog.h>
 #include <assert.h>
+#include <common/pg_prng.h>
 #include <float.h>
+#include <math.h>
 #include <storage/bufmgr.h>
 #include <utils/array.h>
 #include <utils/rel.h>
@@ -18,6 +20,20 @@
 #include "usearch.h"
 #include "utils.h"
 #include "vector.h"
+
+/*
+ * Generate a random level for a new externally stored vector
+ * Rewrite of usearch_newnode_level from usearch, but using a Postgres' rng in <common/pg_prng.h> and its global state
+ * as the source of randomness
+ */
+static uint32 hnsw_generate_new_level(size_t connectivity)
+{
+    // todo:: we can precompute inverse_log_connectivity since it is constant
+    double inverse_log_connectivity = 1.0 / log((double)connectivity);
+    double rand_num = pg_prng_double(&pg_global_prng_state);
+    double level = -1 * log(rand_num) * inverse_log_connectivity;
+    return (uint32)level;
+}
 
 /*
  * Insert a tuple into the index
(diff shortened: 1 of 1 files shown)


ezra-varady:
It looks like the prng header you're using was added to postgres relatively recently, so it breaks older builds. One potential option is to add preprocessor guards to the function to fall back to using C's builtin prng. However I think this solution is less than ideal.

 I believe that the reason the level rng isn't behaving as expected is because the random number generator that is drawn from [here](https://github.com/Ngalstyan4/usearch/blob/9c80e20c1397beed1186d31e31a7962fc6109f81/include/usearch/index.hpp#L2182-L2184) is null initialized [here](https://github.com/Ngalstyan4/usearch/blob/ff268f1759320e885a6ed346e1e3bf4d6af5cda4/include/usearch/index.hpp#L1761). This is probably not an issue if you initialize the index once and make several calls, but because our calling pattern is 1 initialization and then a fixed number of calls to the prng, it's misbehaving. A solution to this bug might be altering the default initialization of this struct member to be seeded with the current time or something similar. This is probably also preferable because it keeps the functionality within the library and would correct the issue at internal call sites as well

Ngalstyan4:
@therealdarkknight, Good work!
You can add a preprocessor guard, as @ezra-varady suggested (see example [here](https://github.com/lanterndata/lantern/blob/40dcf108fda48a1bc67f135da45ef9cec5ef1d4d/src/hnsw/build.c#L27-L31)) and call an appropriately scaled version of `random()` for older versions of postgres.

We actually want this behavior in lantern because we depend on the size of the node to be inserted and this way we can directly calculate the size of the new node fully in lantern

Issue #162, opened 2023-09-19
Add a test that checks index creation on expression


Ngalstyan4:
Test the case when the table does not have a vector column but an hnsw index is create on an expression-generated vector

something like
```
CREATE INDEX ON papers(clip_text(abstract));
```

context:
https://www.postgresqltutorial.com/postgresql-indexes/postgresql-index-on-expression/

dqii:
I wrote this test -- there seems to be a bug though

PR #175, opened 2023-09-24
Infer array dimensions from index expressions

var77:
## Description
When index is being created via expression and `dim` option is not passed we can not infer the array dimensions from the table row. So by this PR we will check if the expression list is not empty when inferring the dimensions, we will evaluate the expression over the first row of the table and try to get the array dimensions from the result.

Relates to #174
Issue #162


dqii (approved):
Clever :)

scripts/extern_defined.sh
```diff
@@ -39,6 +39,8 @@ PG_BIN=$(pg_config --bindir)/postgres
 nm -D  $PG_BIN | grep " T " | awk '{print $3}' | sed -e 's/@.*$//p'
 # global bss symbol in postgres
 nm -D  $PG_BIN | grep " B " | awk '{print $3}' | sed -e 's/@.*$//p'
+# postgres Initialized data (bbs), global symbols
+nm -D  $PG_BIN | grep " D " | awk '{print $3}' | sed -e 's/@.*$//p'
```
Ngalstyan4, line 43: Just curious, were undefined symbols silently passing compilation and that is how you noticed/added this?

src/hnsw/build.c
```diff
@@ -132,7 +134,59 @@ static void BuildCallback(
...
+#if PG_VERSION_NUM >= 120000
+    ExecStoreHeapTuple(tuple, slot, false);
+#else
+    ExecStoreTuple(tuple, slot, InvalidBuffer, false);
+#endif
+    // Set up the tuple for the expression evaluation
+    econtext->ecxt_scantuple = slot;
+
+    // Evaluate the expression for the first row
+    result = ExecEvalExprSwitchContext(exprstate, econtext, &isNull);
```
Ngalstyan4, line 171: what happens when the expression is not constant (for the same row it can return different results)?
What happens if the expression used to create the index with returns an array of different dimension for different rows?
I am pretty sure those cases are covered but let's make sure those are in regression tests and result in sane error messages

src/hnsw/build.c
```diff
@@ -132,7 +134,59 @@ static void BuildCallback(
...
+
+    // Release tuple descriptor
+    ReleaseTupleDesc(tupdesc);
+
+    // Check if the result is not null
+    if(!isNull) {
```
Ngalstyan4, line 177: for this kind of cases, I think it is cleaner to write the code as: 
```
if (error) {
    elog(ERROR, ...);
}

/**/ non-error case with no branching
arr = ..
return n_items;
```

src/hnsw/build.c
```diff
@@ -132,7 +134,59 @@ static void BuildCallback(
...
+        // todo check if datum is array
+        array = DatumGetArrayTypePCopy(result);
+        n_items = ArrayGetNItems(ARR_NDIM(array), ARR_DIMS(array));
+        return n_items;
+    } else {
+        elog(ERROR, "Expression did not result in an array");
```
Ngalstyan4, line 183: Maybe give more context in the error?
e.g.
`Expression used in CREATE INDEX statement did not result in hnsw-index compatible array`

Keep in mind that this is all the user will see when they run an illegal index creation like
`CREATE INDEX ON papers(TO_UPPERCASE(abstract));`. So the error should give them a clue what went wrong.

src/hnsw/build.c
```diff
@@ -161,19 +215,24 @@ static int GetArrayLengthFromHeap(Relation heap, int indexCol)
         return n_items;
     }
 
-    // Get the indexed column out of the row and return it's dimensions
-    datum = heap_getattr(tuple, indexCol, RelationGetDescr(heap), &isNull);
-    if(!isNull) {
-        array = DatumGetArrayTypePCopy(datum);
-        n_items = ArrayGetNItems(ARR_NDIM(array), ARR_DIMS(array));
+    if(indexInfo->ii_Expressions != NULL) {
+        Expr *indexpr_item = lfirst(list_head(indexInfo->ii_Expressions));
```
Ngalstyan4, line 219: Can this list have more expressions? if so, when?
Either case, can you write a comment and relevant invariant check? E.g. if the list will always have a single element, then add an invariant check asserting that the list length is 1

Ngalstyan4 (commented):
Good catch and fix!

I am a little worried about extracting a row from the table and passing it to an expression in index creation but we should be fine with extensive regression testing. We just need to make sure we use the APIs correctly for all kinds of corner cases.

Remaining: add tests
- [x] expression is a non-const function
- [ ] expression returns e.g. 5-lengthed array on the first row, and 6-lengthed array on the second (should be easy to implement if you make the expression be the identity function and use a table with a column that has `real[]` type)
- [ ] add tests from pull #174
- [ ] add pgvector tests (need not implement this functionality. just make sure a sane error is thrown)
- [ ] test when the expression does not even return an array but returns something else
- [ ] test when the expression returns a pgvector array

scripts/extern_defined.sh
var77: They were failing from CI [here](https://github.com/lanterndata/lantern/actions/runs/6290682651/job/17078321222#step:3:634)

src/hnsw/build.c
var77: It can not contain more than 1 element as our hnsw index does not support multi column indexing. Added invariant check for it

src/hnsw/build.c
var77: It is checked when inserting tuples to index, added tests for it

var77:
Currently after creating an index with expression, we are not able to query it using `<->` operator as the operator usage check fails. I have described the issue in [this comment](https://github.com/lanterndata/lantern/pull/175/files#diff-ec8cce8b5c8e24e099bf4bec9c5cbd0a19d9f565846bc60f2e12a8ac3492efdfR91-R102) . Should we fix the issue in this PR or work on it in a different PR?

cc: @Ngalstyan4 @dqii

src/hnsw.c
```diff
@@ -7,6 +7,7 @@
 #include <catalog/namespace.h>
 #include <commands/vacuum.h>
 #include <float.h>
+#include <hnsw/utils.h>
```
Ngalstyan4, line 10: is this our header, included a second time?

src/hnsw/build.c
```diff
@@ -132,7 +134,71 @@ static void BuildCallback(
...
+    // Release tuple descriptor
+    ReleaseTupleDesc(tupdesc);
+
+    Oid oid = get_array_type(array->elemtype);
+
+    if(!OidIsValid(oid)) {
+        // Oid 0 can only be in a case when the datum will be vector type
+        // As postgres will guard that arbitrary data type won't be returned
+        // from the function expression
+        // so we should be safe here to case result into vector in this case
```
Ngalstyan4, line 182: is there any way we could `ldb_invariant` this here?
Will probably need to extract return type info from `expression`

test/sql/hnsw_create_expr.sql
```diff
@@ -13,9 +13,56 @@ BEGIN
...
 $$ LANGUAGE plpgsql IMMUTABLE;
 
+CREATE OR REPLACE FUNCTION int_to_dynamic_binary_real_array(n INT) RETURNS REAL[] AS $$
+DECLARE
+    binary_string TEXT;
+    real_array REAL[] := '{}';
+    i INT;
+    result_length INT;
+BEGIN
+    binary_string := lpad(CAST(n::BIT(3) AS TEXT), 3, '0');
```
Ngalstyan4, line 23: what is the number `3` doing in here?
can you write a comment clarifying what this and the next functions do?

test/sql/hnsw_create_expr.sql
```diff
@@ -13,9 +13,56 @@ BEGIN
...
+
+-- This should result in an error that data type text has no default operator class
+CREATE INDEX ON test_table USING hnsw (int_to_string(id)) WITH (M=2);
+
+-- This should result in error about multicolumn expressions support
+CREATE INDEX ON test_table USING hnsw (int_to_fixed_binary_real_array(id), int_to_dynamic_binary_real_array(id)) WITH (M=2);
+
+-- This currently results in an error about using the operator outside of index
+-- This case should be fixed
+SELECT id FROM test_table ORDER BY int_to_fixed_binary_real_array(id) <-> int_to_fixed_binary_real_array(0) LIMIT 2;
```
Ngalstyan4, line 68: this should be in `hnsw_todos.sql`

Ngalstyan4: Is there any `select` statement on externally created index that works?
e.g. in the example above, if in stead of `int_to_fixed_binary_real_array(0)`, you had a concrete value, would the select work?
If so, add a test of select that works. Otherwise, this is fine.

Ngalstyan4 (approved):
Some nits. Feel free to merge after addressing those

test/sql/hnsw_create_expr.sql
var77: Added comments there

src/hnsw/build.c
var77: Added result oid check with invariant

src/hnsw.c
var77: Removed duplicate include

test/sql/hnsw_create_expr.sql
var77: Moved the failing test to `hnsw_todo`, could not find any order by that will work for now, will check this issue on a separate PR.

PR #179, opened 2023-09-25
Support expression-based index operator checks

dqii:


Ngalstyan4:
@var77 , it seems code coverage does not run even when just benchmarking fails. 
Can we make it so that code coverage runs as long as build succeeds and tests pass?

var77:
@Ngalstyan4 I think there's an issue with codecov not running the checks after we have changed the repo name. Can you try to reinstall the Codecov app and update codecov token in action secrets? Maybe that will fix the issue

var77 (approved):

PR #183, opened 2023-09-26
Homebrew formula release

var77:
Added release action to update homebrew formula version.
Added ability to specify pg_config from `PG_CONFIG` env variable on `cmake/FindPostgreSQL.cmake`, so we can be sure that homebrew formula will compile Lantern against the same postgres version it has found.


Ngalstyan4 (approved):
@var77, Merge this once the brew tap is ready

Issue #148, opened 2023-09-16
Rewrite distance operator `<->` into the specific distance function if existing index won't trigger


Ngalstyan4:
Currently we properly detect and fail the query in cases when the distance operator `<->` is used outside of index-search context

We only check for existence of the index, however, and we do not check whether the index will actually be triggered.
This means that on small tables search queries fill fail unless we set `enable_seqscan=off`. This is bad UX.

We can instead detect that index is not used and rewrite the query to trigger the right distance operator.

gnehzza:
I'm currently working on this issue, to avoid someone else overlapping with me.

PR #185, opened 2023-09-28
Changes to address issue 148

gnehzza:
Changes to address this: https://github.com/lanterndata/lantern/issues/148


Ngalstyan4:
Feature discussed and implemented in another PR. See #189

PR #192, opened 2023-10-05
Add parallel tests

ezra-varady:
This PR adds support for a set of parallel tests to ensure that lantern behaves properly under a more realistic workload. At the moment they are not especially interesting, just a PoC, but this includes the changes to support the concept more generally. I have for the time being added these tests to `make test` to see how they'll interact with CI on this PR. They should eventually be restricted to `make test-parallel` @var77 if you have ideas about how best to include them in the github actions workflow any advice would be awesome. 

Some notes:
- To run the new tests alone run `make test-parallel` (currently they'll also run if you run `make test` but I'd like to change this before merging
- The schedule for the new tests is `test/parallel_schedule.txt` 
- The tests themselves live in `test/parallel` they have the same layout as the existing regression tests
- At the moment these tests only bring in sift10k do a few inserts and selects and check some obvious invariants at the end if anyone has suggestions for additional tests/invariants to add please let me know and I will include them


test/parallel/parallel_test_runner.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+# Get current test file name
+TESTFILE_NAME=${PGAPPNAME##pg_regress/}
+# Set different name for each test database
+# As pg_regress does not support cleaning db after each test
```
var77, line 6: I think we can remove this comments

test/parallel/parallel_test_runner.sh
```diff
@@ -0,0 +1,52 @@
...
+# Change directory to sql so sql imports will work correctly
+cd sql/
+# install lantern extension
+if [[ "$TESTFILE_NAME" =~ ^begin ]]; then
+    psql "$@" -U ${DB_USER} -d postgres -v ECHO=none -q -c "CREATE DATABASE ${TEST_CASE_DB};" 2>/dev/null
+    psql "$@" -U ${DB_USER} -d ${TEST_CASE_DB} -v ECHO=none -q -c "SET client_min_messages=error; CREATE EXTENSION lantern;" 2>/dev/null
+    #psql "$@" -U ${DB_USER} -d ${TEST_CASE_DB} -v ECHO=none -q -f utils/common.sql 2>/dev/null
+fi
+
+# Exclude debug/inconsistent output from psql
```
var77, line 37: is it possible to create some `common_runner` file wrap this into a function and use that in both runners? Also if it's an option we can separate all the common parts in these scripts

test/parallel/parallel_test_runner.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Drop db after each test on exit signal
+function drop_db {
+  cat <<EOF | psql "$@" -U ${DB_USER} -d postgres -v ECHO=none >/dev/null 2>&1
+    SET client_min_messages=ERROR;
+    DROP DATABASE "${TEST_CASE_DB}";
+EOF
+}
+
+if [[ "$TESTFILE_NAME" =~ ^end ]]; then
```
var77, line 23: can we document somewhere. that `begin.sql` and `end.sql` are treated differently in the runner? We can just add comments in this file

test/parallel/sql/select.sql
```diff
@@ -0,0 +1,8 @@
+SELECT v AS v1111 FROM sift_base10k WHERE id = 1111 \gset
+SELECT v AS v2222 FROM sift_base10k WHERE id = 2222 \gset
+SELECT v AS v3333 FROM sift_base10k WHERE id = 3333 \gset
+SELECT v AS v4444 FROM sift_base10k WHERE id = 4444 \gset
+SELECT id FROM sift_base10k ORDER BY  v <-> :'v1111'  ASC LIMIT 1;
+SELECT id FROM sift_base10k ORDER BY  v <-> :'v2222'  ASC LIMIT 1;
```
var77, line 6: can we also have `EXPLAIN` case here with enable_seqscan=false , as after the merge of [this](https://github.com/lanterndata/lantern/pull/189) PR it won't be obvious that this was an index scan or not.

var77 (commented):
Look good, great changes! Added some small comments. For the CI we may modify [run-tests-linux.sh](https://github.com/lanterndata/lantern/blob/main/ci/scripts/run-tests-linux.sh) and [run-tests-mac.sh](https://github.com/lanterndata/lantern/blob/main/ci/scripts/run-tests-mac.sh) files. to run the parallel tests as well

test/parallel/parallel_test_runner.sh
ezra-varady: Ah yeah, I had initially split it out so I could test more easily, I'll combine these

test/parallel/parallel_test_runner.sh
ezra-varady: sounds good, I'll add comments in the schedule and the files themselves too

var77 (approved):

scripts/run_all_tests.sh
```diff
@@ -67,13 +67,33 @@ fi
...
+fi
 if [ -n "$FILTER" ]; then
-    if [[ "$pgvector_installed" == "1" ]]; then
-        TEST_FILES=$(cat schedule.txt | grep -E '^(test:|test_pgvector:)' | sed -E -e 's/^test:|test_pgvector://' | tr " " "\n" | sed -e '/^$/d')
+    if [ "$PARALLEL" -eq 1 ]; then
+    	TEST_FILES=$(cat $SCHEDULE | grep -E '^(test:|test_begin:|test_end:)' | sed -E -e 's/^test:|test_begin:|test_end://' | tr " " "\n" | sed -e '/^$/d')
     else
-        TEST_FILES=$(cat schedule.txt | grep '^test:' | sed -e 's/^test://' | tr " " "\n" | sed -e '/^$/d')
+	    if [[ "$pgvector_installed" == "1" ]]; then
+		TEST_FILES=$(cat $SCHEDULE | grep -E '^(test:|test_pgvector:)' | sed -E -e 's/^test:|test_pgvector://' | tr " " "\n" | sed -e '/^$/d')
```
Ngalstyan4, line 93: Unrelated to this PR, but still important: Do we currently run pgvector compat tests anywhere in ci/cd or release pipeline?

test/parallel/sql/utils/random_array.sql
```diff
@@ -0,0 +1,11 @@
+CREATE OR REPLACE FUNCTION random_int_array(dim integer, min integer, max integer) RETURNS integer[] AS $BODY$
```
Ngalstyan4, line 1: I feel like this would be useful in regular tests as well. Should be in the utils for regular tests.

Ngalstyan4 (approved):
a couple nits. otherwise ready to merge. Well done!

scripts/run_all_tests.sh
var77: Yes we are running them in the pipeline, as we're installing the pgvector there

test/test_runner.sh
```diff
@@ -20,16 +27,34 @@ function drop_db {
...
-psql "$@" -U ${DB_USER} -d postgres -v ECHO=none -q -c "DROP DATABASE IF EXISTS ${TEST_CASE_DB};" 2>/dev/null
-psql "$@" -U ${DB_USER} -d postgres -v ECHO=none -q -c "CREATE DATABASE ${TEST_CASE_DB};" 2>/dev/null
-psql "$@" -U ${DB_USER} -d ${TEST_CASE_DB} -v ECHO=none -q -c "SET client_min_messages=error; CREATE EXTENSION lantern;" 2>/dev/null
-psql "$@" -U ${DB_USER} -d ${TEST_CASE_DB} -v ECHO=none -q -f utils/common.sql 2>/dev/null
+# if tests are parallel we only do this for the begin tests as we won't be dropping the database until the end
+# begin will handle initialization specific to the tests but expects the database already exists
+if [ "$PARALLEL" -eq 0 ]; then
+    psql "$@" -U ${DB_USER} -d postgres -v ECHO=none -q -c "DROP DATABASE IF EXISTS ${TEST_CASE_DB};" 2>/dev/null
+    db_init
+    psql "$@" -U ${DB_USER} -d ${TEST_CASE_DB} -v ECHO=none -q -f utils/common.sql 2>/dev/null
```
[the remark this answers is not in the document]
ezra-varady: good catch, I'll rework this

ezra-varady:
merged the two branches in the test runner and made copies of a couple utilities

PR #200, opened 2023-10-11
Add warnings when the index size exceeds work_mem

ezra-varady:
This adds checks for the `work_mem` and `maintenance_work_mem` GUC variables. Checks for both are fairly uncommon in the upstream extensions. It seems like `maintenance_work_mem` is checked more often than `work_mem` this may be because it's greater size makes it less prohibitive to respect. Neither is actually enforced by postgres. I'll list some examples below 

`maintenance_work_mem` is referenced more infrequently but its use is pretty straightforward when it is
in `src/backend/access/gin/gininsert.c` in the build callback function `maintenance_work_mem` is checked
```
if (buildstate->accum.allocatedMemory >= (Size) maintenance_work_mem * 1024L)
```

similarly nbtree checks it in its parallel build functions `src/backend/access/nbtree/nbtsort.c`
```
 sortmem = maintenance_work_mem / btshared->scantuplesortstates;
 _bt_parallel_scan_and_sort(btspool, btspool2, btshared, sharedsort,
                                               sharedsort2, sortmem, false);
```
It is also checked in `src/backend/commands/vacuumparallel.c`. It is never checked in contrib. I think the `bloom` in the earlier listing refers to an internal bloomfilter, not the extension. Notably though pgvector does have a [check](https://github.com/pgvector/pgvector/blob/e630efd195c563496c3550abb1817303586ee46d/src/hnswbuild.c#L386-L400)

`work_mem` is also checked very infrequently although within the optimizer/executor there are a number of checks. 
in `src/backend/access/gin/ginfast.c` it gets checked
```
workMemory = work_mem;
...
/*
* Is it time to flush memory to disk?  Flush if we are at the end of
* the pending list, or if we have a full row and memory is getting
* full.
*/
if (GinPageGetOpaque(page)->rightlink == InvalidBlockNumber ||
    (GinPageHasFullRow(page) &&
    (accum.allocatedMemory >= workMemory * 1024L)))
```
it gets checked in `src/backend/access/nbtree/nbtpage.c` as well, albeit in a function that is only called during vacuums
```
maxbufsize = (work_mem * 1024L) / sizeof(BTPendingFSM);
maxbufsize = Min(maxbufsize, INT_MAX);
maxbufsize = Min(maxbufsize, MaxAllocSize / sizeof(BTPendingFSM));
/* Stay sane with small work_mem */
maxbufsize = Max(maxbufsize, vstate->bufsize);
vstate->maxbufsize = maxbufsize;
```
I have however found the following calling pattern in several places including `contrib/tablefunc/tablefunc.c`, `contrib/dblink/dblink.c`, `contrib/adminpack/adminpack.c` and also [pgvector](https://github.com/pgvector/pgvector/blob/e630efd195c563496c3550abb1817303586ee46d/src/ivfscan.c#L228) (albeit only in ivfscans)
```
tuplestore_begin_heap(random_access, false, work_mem);
```
this seems to be a data structure that holds tuples to be returned by a scan. It doesn't account for memory allocated elsewhere though. Overall the lack of enforcement seems to make checking these values somewhat uncommon. I think it makes sense to enforce `maintenance_work_mem` because building an index is relatively infrequent, but maybe enforcing runtime checks for `work_mem` is overkill

### DIFF
--- a/src/hnsw/build.c
+++ b/src/hnsw/build.c
@@ -9,6 +9,8 @@
 #include <catalog/pg_type.h>
 #include <executor/executor.h>
 #include <funcapi.h>
+#include <math.h>
+#include <miscadmin.h>
 #include <nodes/execnodes.h>
 #include <storage/bufmgr.h>
 #include <utils/array.h>
@@ -59,7 +61,7 @@
 #define UpdateProgress(index, val) ((void)val)
 #endif
 
-static void AddTupleToUsearchIndex(ItemPointer tid, Datum *values, HnswBuildState *buildstate)
+static void AddTupleToUsearchIndex(ItemPointer tid, Datum *values, HnswBuildState *buildstate, Relation index)
 {
     /* Detoast once for all calls */
     usearch_error_t       error = NULL;
@@ -92,6 +94,15 @@ static void AddTupleToUsearchIndex(ItemPointer tid, Datum *values, HnswBuildStat
     if(buildstate->usearch_index != NULL) {
         size_t capacity = usearch_capacity(buildstate->usearch_index, &error);
         if(capacity == usearch_size(buildstate->usearch_index, &error)) {
+            double             M = ldb_HnswGetM(index);
+            double             mL = 1 / log(M);
+            usearch_metadata_t meta = usearch_metadata(buildstate->usearch_index, &error);
+            uint32             node_size = UsearchNodeBytes(&meta, meta.dimensions * sizeof(float), (int)(mL + .5));
+            if(2 * usearch_size(buildstate->usearch_index, &error) * node_size
+               >= (size_t)maintenance_work_mem * 1024L) {
+                usearch_free(buildstate->usearch_index, &error);
+                elog(ERROR, "index size exceeded maintenance_work_mem during index construction");
+            }
             usearch_reserve(buildstate->usearch_index, 2 * capacity, &error);
             assert(error == NULL);
         }
@@ -130,7 +141,7 @@ static void BuildCallback(
 
     // todo:: the argument values is assumed to be a real[] or vector (they have the same layout)
     // do proper type checking instead of this assumption and test int int arrays and others
-    AddTupleToUsearchIndex(tid, values, buildstate);
+    AddTupleToUsearchIndex(tid, values, buildstate, index);
 
     /* Reset memory context */
     MemoryContextSwitchTo(oldCtx);
@@ -452,6 +463,14 @@ static void BuildIndex(
             estimated_row_count = offset * numBlocks;
             // Unlock and release buffer
             UnlockReleaseBuffer(buffer);
+        }
+        double             M = ldb_HnswGetM(index);
+        double             mL = 1 / log(M);
+        usearch_metadata_t meta = usearch_metadata(buildstate->usearch_index, &error);
+        uint32             node_size = UsearchNodeBytes(&meta, opts.dimensions * sizeof(float), (int)(mL + .5));
+        // accuracy could be improved by not rounding mL, but otherwise this will never be fully accurate
+        if(node_size * estimated_row_count > maintenance_work_mem * 1024L) {
+            elog(ERROR, "index size exceeded maintenance_work_mem during index construction");
         }
         usearch_reserve(buildstate->usearch_index, estimated_row_count, &error);
         if(error != NULL) {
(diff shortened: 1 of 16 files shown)


src/hnsw/build.c
```diff
@@ -91,6 +93,15 @@ static void AddTupleToUsearchIndex(ItemPointer tid, Datum *values, HnswBuildStat
     if(buildstate->usearch_index != NULL) {
         size_t capacity = usearch_capacity(buildstate->usearch_index, &error);
         if(capacity == usearch_size(buildstate->usearch_index, &error)) {
+            double             M = ldb_HnswGetM(index);
+            double             mL = 1 / log(M);
+            usearch_metadata_t meta = usearch_metadata(buildstate->usearch_index, &error);
+            uint32             node_size = UsearchNodeBytes(&meta, meta.dimensions * sizeof(float), (int)(mL + .5));
```
Ngalstyan4, line 99: where is the `mL+.5` term coming from?
I assume this is the expected node size, given the distribution in the paper. Could you write a note in the comment around it?

Ngalstyan4, line 99: Note that once #19 is merged, `sizeof(float)` will. no longer be correct, since with type casts vector elements can be smaller than `sizeof(float)`. Could you add a note (`todo:: update sizeof(float) to correct vector size once #19 is merged`)so we do not overlook it later?

src/hnsw/build.c
```diff
@@ -91,6 +93,15 @@ static void AddTupleToUsearchIndex(ItemPointer tid, Datum *values, HnswBuildStat
...
+            if(2 * usearch_size(buildstate->usearch_index, &error) * node_size
+               >= (size_t)maintenance_work_mem * 1024L) {
+                usearch_free(buildstate->usearch_index, &error);
+                elog(ERROR, "index size exceeded maintenance_work_mem during index construction");
```
Ngalstyan4, line 103: For now, let's leave this as a warning. 
We are carrying out fault tolerance tests now and we should throw errors like this once we are sure we handle them well.

src/hnsw/build.c
```diff
@@ -452,6 +463,14 @@ static void BuildIndex(
             // Unlock and release buffer
             UnlockReleaseBuffer(buffer);
         }
+        double             M = ldb_HnswGetM(index);
+        double             mL = 1 / log(M);
+        usearch_metadata_t meta = usearch_metadata(buildstate->usearch_index, &error);
+        uint32             node_size = UsearchNodeBytes(&meta, opts.dimensions * sizeof(float), (int)(mL + .5));
+        // accuracy could be improved by not rounding mL, but otherwise this will never be fully accurate
+        if(node_size * estimated_row_count > maintenance_work_mem * 1024L) {
+            elog(ERROR, "index size exceeded maintenance_work_mem during index construction");
```
Ngalstyan4, line 472: same as above, about the error.

src/hnsw/build.c
```diff
@@ -452,6 +463,14 @@ static void BuildIndex(
             // Unlock and release buffer
             UnlockReleaseBuffer(buffer);
         }
+        double             M = ldb_HnswGetM(index);
+        double             mL = 1 / log(M);
+        usearch_metadata_t meta = usearch_metadata(buildstate->usearch_index, &error);
```
Ngalstyan4, line 468: Is this the exact same code as above?
Can we factor it out into a `Check__` function?

Ngalstyan4 (commented):
Looks good!
- Could you reduce code duplication?
It seems very similar code blocks are added in `insert`, `scan` and `build`. If possible, we should move these checks to a helper function

- It seems the general strategy in this PR is to instrument memory allocation points with a conditional check for memory limits.
    1. Could we use `CurrentMemoryContext` to get a more accurate measure of used memory?
(I think at any given point in time we can ask postgres to tell us how much memory is allocated at `CurrentMemoryContext`) 

    2. Is there a way to cap maximum memory allowed in a memory context when we create the memory context? We create such contexts for builds, inserts etc. And if we can enforce memory constraints there, we will have less code to maintain

ezra-varady:
Rebased onto main. Broke out code to check memory use into a helper function in utils.c. I added code to include memory allocated in the current context and it's children in these calculations for versions of postgres greater than 12. Before pg13 there's not a clear way to check how much memory has been allocated by a single context. The call site in the node retriever will get called a lot. It may be worth trying to optimize a bit around this

Ngalstyan4:
Looks good! moved to #205. Will change one test, check code coverage and merge from there

PR #211, opened 2023-10-21
WAL/blockmaps: continue blockmaps initialization after failure

medvied:
ContinueBlockMapGroupInitialization() description has all the details about how it's done and how different kinds of failures are handled.


medvied:
The tests don't fail with #217, so this patch should be merged after #217 is merged.

PR #214, opened 2023-10-25
Figuring out why tests fail on main

Ngalstyan4:
Currently CI runs fail once a PR is merged, even though all tests pass while the change is an open PR.

The culprit is the recently added version update script. It seems somehow repo tags exist while the CI/CD is running on a pr but the repo tags are missing once CI/CD is running on the merged commit on main.

[This](https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches) made me think that the checkout action by default does not pull tag info and this PR attempts to fix it.

@var77, Do you think this the issue? Do you think my changes would fix it?
Is there a way to test this before we merge?

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -51,7 +51,9 @@ jobs:
           sudo pip install GitPython &&\
           # Start postgres
           sudo su postgres -c "PG_VERSION=$PG_VERSION RUN_TESTS=0 ./ci/scripts/run-tests-linux.sh" && \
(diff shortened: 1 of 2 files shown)


var77:
That is strange, I had added a line to fetch from remote in python script before, but seems it's not helping here. I can test this in my fork later today and let you know

var77:
@Ngalstyan4 in my fork [here](https://github.com/var77/lantern/actions/runs/6636834860) it is passing the tests in main

var77 (approved):

PR #218, opened 2023-10-26
Implement failure points

medvied:
This patch adds API to trigger execution of C code from SQL to test corner cases.

`test/sql/hnsw_failure_point.sql` has an example of how to trigger a process crash using failure points and how to see that a space leak happens if a crash happens after a block is allocated, but before a record for the block is added to the index during blockmaps creation.


src/hnsw/failure_point.c
```diff
@@ -0,0 +1,75 @@
...
+             state->name,
+             state->remaining,
+             func,
+             name,
+             dont_trigger_first_nr);
+    }
+    *state = (struct failure_point_state){
+        .enabled = true,
+        .func = func,
+        .name = name,
```
Ngalstyan4, line 48: Do `name` and `func` live long enough for when it is used from `state`?

Both  go out of scope and are freed as soon as the SQL function returns. It is allocated by the function context. 

so, at that point the global `state.name` is dangling.
If this is the case, you should probably copy the strings. We do not care about performance for these, it should be fine

If you need the name after the

medvied: Good catch. Done.

I had a choice between using `malloc()` directly, using `char func[0x100];` for the `state->func` and using a custom PostgreSQL allocator for this file. `char func[0x100];` won.

PR #219, opened 2023-10-31
Rename hnsw.init_k to lantern_hnsw.init_k

var77:
Rename `hnsw.init_k` parameter to `lantern_hnsw.init_k` to match the naming convention we have for `lantern_hnsw.ef_search` parameter

### DIFF
--- a/src/hnsw/options.c
+++ b/src/hnsw/options.c
@@ -222,34 +222,7 @@ void _PG_init(void)
                             NULL,
                             NULL);
 
-    DefineCustomIntVariable("hnsw.ef_search",
-                            "Expansion factor to use during vector search in a scan",
-                            "Valid values are in range [1, 400]",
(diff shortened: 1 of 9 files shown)


Ngalstyan4:
should this only be done when `pgvector` is also installed?
we only use `lantern_hnsw` if `hnsw` is already taken by pgvector

var77:
I did this change to match the naming with the other variable we have `lantern_hnsw.ef_search` . As that also has the prefix `lantern_`, we can decide to use hnsw or lantern_hnsw (or use lantern_hnsw only when pgvector is installed) what do you think?

Ngalstyan4:
hm, `lantern_hnsw` was supposed to only be for pgvector compatibility.
hnsw must exist otherwise. (`lantern_hnsw` can exist otherwise as well, but `hnsw` MUST exist)
So, this should be changed accordingly

PR #231, opened 2023-11-30
Fix wrong memory context for allocation in op_rewrite.c

var77:
The nodes were being allocated in `MessageContext` which doesn't live long enough and in some cases it was causing segfault when postgres was trying to access the nodes. Per [source-code](https://doxygen.postgresql.org/pquery_8c.html#a53a9d1887eabd76cd7f9e3be4b5ac62c) the plan tree is being allocated in `PortalContext`, so we also should allocate the overwritten nodes in `PortalContext`


Ngalstyan4:
We should check and see whether the same change must be done [here](https://github.com/lanterndata/lantern/commit/1c2ff533c19c43fb186a4411dac16c666ed44a41#diff-333614cfbeb90cb676f6b37336d24e3e771edaa6fceaf78a16cfff4368b83c3aR206) as well or no.

In there we a new Function Expression node.

PR #233, opened 2023-12-01
Added dockerfile with postgres debug build

var77:
This dockerfile can help to better debug the database crashes as it has postgres built with debug symbols.
In the top of the Dockerfile you can find the instructions on how to build and run the image to use gdb


Dockerfile.debug
```diff
@@ -0,0 +1,41 @@
...
+ARG VERSION=15.3
+
+RUN apt update && apt install wget git make build-essential libreadline8 libreadline-dev zlib1g zlib1g-dev cmake gdb tmux vim sudo -y
+
+# Build and install postgres
+RUN cd /root && wget https://ftp.postgresql.org/pub/source/v${VERSION}/postgresql-${VERSION}.tar.bz2 && \
+    tar xf postgresql-${VERSION}.tar.bz2 && \
+    rm -rf postgresql-${VERSION}.tar.bz2 && \
+    cd postgresql-${VERSION} && ./configure --enable-debug && \
+    make && make install
```
Ngalstyan4, line 18: Can we build and push up to this point to dockerHUB?
Would be annoying to build postgres from source.

Alternatively, if there is a docker image with postgres installed from source, we could use that

var77: Yes I can push that to dockerhub for 15 version and use that image

Ngalstyan4:
Are there any downsides to using this image? 
Should we just update [Dockerfile.dev](https://github.com/lanterndata/lantern/blob/main/Dockerfile.dev) with this, so in development we always get full stack traces into postgres.

Ngalstyan4 (approved):

PR #238, opened 2023-12-03
Add C client tests to catch op_rewriting segfault

var77:
Added tests with C libpq client to catch the segfault happening on operator rewrite hook.

To add new tests you need to do the following:
1. Create test file under `test/c/`
2. Include test file in `test/runner.c`
3. Add test case to `test_cases[]` in `runner.c`

### DIFF
--- a/ci/scripts/run-tests-linux.sh
+++ b/ci/scripts/run-tests-linux.sh
@@ -27,6 +27,7 @@ function run_db_tests(){
   then
     cd $WORKDIR/build && \
     make test && \
+    make test-c
     killall postgres && \
     gcovr -r $WORKDIR/src/ --object-directory $WORKDIR/build/ --xml /tmp/coverage.xml
   fi
--- a/ci/scripts/run-tests-linux.sh
+++ b/ci/scripts/run-tests-linux.sh
@@ -27,7 +27,7 @@ function run_db_tests(){
   then
     cd $WORKDIR/build && \
     make test && \
-    make test-c
+    make test-client
     killall postgres && \
     gcovr -r $WORKDIR/src/ --object-directory $WORKDIR/build/ --xml /tmp/coverage.xml
   fi
--- a/test/c/runner.c
+++ b/test/c/runner.c
@@ -83,6 +83,17 @@ int create_extension(PGconn *conn)
     return 0;
 }
 
+const char *getenv_or_default(const char *env_name, const char *default_val)
+{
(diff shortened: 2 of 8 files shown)


ci/scripts/run-tests-linux.sh
```diff
@@ -27,6 +27,7 @@ function run_db_tests(){
   then
     cd $WORKDIR/build && \
     make test && \
+    make test-client
```
Ngalstyan4, line 30: If it is trivial to run this on mac as well, let's do that. But otherwise it is fine as is.

test/c/runner.c
```diff
@@ -0,0 +1,162 @@
...
+    free(full_statement);
+
+    if(PQresultStatus(res) != PGRES_COMMAND_OK) {
+        fprintf(stderr, "Failed to drop test db: %s\n", PQerrorMessage(root_conn));
+        PQclear(res);
+        return 1;
+    }
+
+    statement = "CREATE DATABASE ";
+    full_statement = malloc(strlen(statement) + strlen(test_db_name));
```
Ngalstyan4, line 59: Should there be a `+1` in malloc size here for `\n` character in sprintf below?

test/c/test_op_rewrite.c
```diff
@@ -0,0 +1,54 @@
+#ifndef TEST_OP_REWRITE_H
```
Ngalstyan4, line 1: not needed in a C file

Ngalstyan4 (commented):
Looks good!

test/c/runner.c
var77: True, forgot about nullbyte.

ci/scripts/run-tests-linux.sh
var77: Added to mac tests as well.

test/c/runner.c
```diff
@@ -19,7 +19,7 @@ struct TestCase
 PGconn *connect_database(
     const char *db_host, const char *db_port, const char *db_user, const char *db_password, const char *db_name)
 {
-    const int const_db_uri_chars = strlen("host= port= user= dbname= sslmode=disable password=");
+    const int const_db_uri_chars = strlen("host= port= user= dbname= sslmode=disable password=") + 1;
```
Ngalstyan4, line 22: +1 in malloc, like in other cases

Ngalstyan4 (approved):
looks good. feel free to merge!

PR #244, opened 2023-12-11
Rename generic operator.

var77:
Rename generic operator `<->` to `<?>` and fix update script for `0.0.9->0.0.10`


sql/updates/0.0.9--0.0.10.sql
```diff
@@ -1,5 +1,177 @@
...
+		WHERE typname = 'vector'
+	) INTO pgvector_exists;
+	
+	am_name := 'hnsw';
+		
+
+	IF pgvector_exists THEN
+		am_name := 'lantern_hnsw';
+    -- these go for good.
+    DROP OPERATOR CLASS IF EXISTS dist_vec_hamming_ops USING hnsw CASCADE;
```
Ngalstyan4, line 80: should this be `USING lantern_hnsw`?

Ngalstyan4: or use `am_name`? It is the code I had written in the last PR and I think I forgot to consider the changed naming of our AM when pgvector is present

test/expected/hnsw_todo.out
```diff
@@ -38,14 +38,15 @@ ERROR:  relation "small_world_l2_vector_int_idx" does not exist at character 41
...
+                              QUERY PLAN                               
+-----------------------------------------------------------------------
  Limit
-   ->  Sort
-         Sort Key: (public.l2sq_dist(vector_int, '{0,1,0}'::integer[]))
-         ->  Seq Scan on small_world_l2
-(4 rows)
+   ->  Result
+         ->  Sort
+               Sort Key: (l2sq_dist(vector_int, '{0,1,0}'::integer[]))
```
Ngalstyan4, line 47: hm, do you know why before the Sort Key function was schema-qualified (`public.l2sq_dist`) while now it is not (`l2sq_dist`) ?

Ngalstyan4 (approved):
Looks good!

sql/updates/0.0.9--0.0.10.sql
var77: True, that should be lantern_hnsw;

test/expected/hnsw_todo.out
var77: Actually that was added after the latest release, before that it was without public if I remember correctly, not sure why but one change that I did to that operator was adding it for integer type as well

PR #247, opened 2023-12-12
Add fallback mechanism for external index when reindexing

var77:
## Issue
When creating external index using `_experimental_index_path` option without passing any other options we are getting error on `REINDEX` operation if the index file will not exist anymore. The ideal solution will be to trigger external function in that case and create a new external index using multicore approach.

## Description
With this PR I have added a fallback mechanism, which will check if we have provided index params required for index creation (`ef`, `ef_search`, `ef_construction`) it will use them and create index locally like regular index. This will be slow compared to external index but at least it will not fail reindexing.
I have also added checks to verify that provided options are matching the ones defined in the index file header.
I have removed the default option values from `options.c` to be able to check if they are passed or not. This will not break any existing logic as we don't rely on them anyway by using our helper functions (`ldb_HnswGetM`, `ldb_HnswGetEfConstruction`, `ldb_HnswGetEf`)

### DIFF
--- a/src/hnsw/build.c
+++ b/src/hnsw/build.c
@@ -386,6 +386,93 @@ static void FreeBuildState(HnswBuildState *buildstate)
 }
 
 /*
+ * This function will be called when an external file will be passed (_experimental_index_file_path param)
+ * we should check if user has also passed index params verify that it matches with the ones defined in index file
+ * header With this we ensure that on REINDEX operation, if we use that params to construct the index
+ * it will match
+ */
+static void ValidateIndexFileParams(Relation index, usearch_init_options_t *opts)
+{
+    ldb_HnswOptions *index_options = (ldb_HnswOptions *)index->rd_options;
+    assert(index_options != NULL);
+
+    if(index_options->ef && (size_t)index_options->ef != opts->expansion_search) {
+        elog(ERROR,
+             "Index option 'ef'(%u) does not match with the index file option 'ef'(%lu)",
+             index_options->ef,
+             opts->expansion_search);
+    }
+
+    if(index_options->ef_construction && (size_t)index_options->ef_construction != opts->expansion_add) {
+        elog(ERROR,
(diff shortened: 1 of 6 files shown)


var77:
I will reimplement this one after new function will be created for external index reindexing

src/hnsw/build.c
```diff
@@ -385,6 +385,93 @@ static void FreeBuildState(HnswBuildState *buildstate)
...
+ * it will match
+ */
+static void ValidateIndexFileParams(Relation index, usearch_init_options_t *opts)
+{
+    ldb_HnswOptions *index_options = (ldb_HnswOptions *)index->rd_options;
+    assert(index_options != NULL);
+
+    if(index_options->ef && (size_t)index_options->ef != opts->expansion_search) {
+        elog(ERROR,
+             "Index option 'ef'(%u) does not match with the index file option 'ef'(%lu)",
```
Ngalstyan4, line 401: ...does not match the 'ef' parameter with which the external file index was created

src/hnsw/options.c
```diff
@@ -178,7 +178,7 @@ void _PG_init(void)
     add_int_reloption(ldb_hnsw_index_withopts,
                       "m",
                       "HNSW M hyperparameter",
-                      HNSW_DEFAULT_M,
```
Ngalstyan4, line 181: My understanding is you changed the default values to 0 (which is an invalid value of all of these), to later require explicit passing of these parameters when creating an index from a file path. Is this the case?
Does this affect the printed index parameters when a regular `hnsw` index is created with default parameters and when you run `\di` or `\d`on the table to see all the indexes and their creation parameters?

Ngalstyan4 (approved):

src/hnsw/options.c
var77: Actually I have changed them to be able to determine if they were passed or not via comparing to NULL values, but think I will close this PR and throw and error if the file won't exist asking to use `lantern_reindex_external_index()` function from this [PR](https://github.com/lanterndata/lantern/pull/250).

Issue #239, opened 2023-12-04
Lantern binary versioning


Ngalstyan4:
Currently, if one installs a newer version of `lantern.so` binary without running `ALTER EXTENSION ... UPDATE`, we cannot detect it and using lantern indexes or other functions may corrupt the database.

This is handled in the cloud by making sure that we always run `ALTER EXTENSION` after updating the binary. But we should make sure there is an additional guard in the database itself to prevent this case from happening in all environments.

To do that, we should create an autogenerated `config.h` file, that will have database version and build hash as `global constant`s in it. Then, in `_PG_init`, we can check that SQL layer and the binary agree on versions and can refuse to enable the extension otherwise.

PR #249, opened 2023-12-17
Add binary versioning

ezra-varady:
This addresses #239. It generates a version header that gets included in the binary and compared to the SQL API version. If they do not match it will refuse to interact with the index. Unfortunately the system catalog does not make extension information in C afaict, so we execute an actual query to get the SQL API version. The result is cached to minimize the overhead but this means that once the mismatch has been resolved the connection must be reset. I've tested it locally and it behaves correctly.

I also added a new class of tests to check that errors are generated during update tests. I don't love the setup, so if anyone has ideas please let me know. I added a new class of tests, `test-misc` that are intended for use in situations like this where we want to call a single test in a very specific instance. I couldn't reuse the parallel test because the expected output has to be different which was unfortunate. In addition to this I replaced various duplicated utils with symlinks


ezra-varady:
it looks like benchmark is maybe checking out a tag without altering the version. The misc tests can't be used without bumping the version which is a bit annoying, I'll look into this tomorrow. Not sure what the deal with the missing gettext header is on osx either, it looks like it's installed?

scripts/test_updates.py
```diff
@@ -10,6 +10,40 @@
...
 
+class Version:
+    def __init__(self, version: str):
+        self.version_numbers = [int(n) for n in version.split('.')]
+    def __lt__(self, other):
+        for i, v in enumerate(self.version_numbers):
+            if v < other.version_numbers[i]:
+                return True
+        return False
+    def __le__(self, other):
```
Ngalstyan4, line 21: Can the rest be implemented in terms of `__lt__` to avoid repetition?

CMakeLists.txt
```diff
@@ -203,6 +203,15 @@ add_custom_command(
 
 add_custom_target(phony_always_runs ALL DEPENDS ${CMAKE_BINARY_DIR}/${_script_file})
 
+add_custom_target(
+    generate_version_header
+    COMMAND ${CMAKE_SOURCE_DIR}/scripts/version_header.sh ${LANTERNDB_VERSION} ${CMAKE_BINARY_DIR}
```
Ngalstyan4, line 208: `cmake` has a built-in function for this kind of stuff (`configure_file`).
Is there a reason we are not using that?
(See `CONTROL_TEMPLATE` in this CmakeLists.txt for example usage)

src/hnsw/utils.c
```diff
@@ -103,3 +109,65 @@ float4 *ToFloat4Array(ArrayType *arr)
...
+        text       *version_text;
+
+        if(SPI_connect() != SPI_OK_CONNECT) {
+            elog(ERROR, "could not connect to executor to check binary version");
+        }
+
+        query = "SELECT extversion FROM pg_extension WHERE extname = 'lantern'";
+
+        // Execute the query to figure out what version of lantern is in use in SQL
+        // todo: it would be nice to get this without actually executing a query
```
Ngalstyan4, line 136: it is fine to run the query, no worries!

src/hnsw/utils.h
```diff
@@ -6,11 +6,15 @@
 #include "options.h"
 #include "usearch.h"
 
+extern bool versions_match;
```
Ngalstyan4, line 9: do these need to be exported? Isn't the `VersionMatch()` the only API used externally?

ezra-varady:
I haven't been able to recreate the issue with the benchmark runner locally in the dev container. I'm not sure what's going on exactly since this error shouldn't be possible without an update and none is done

ezra-varady:
It looks like maybe it was a caching issue? The test has started passing since I added and then removed the upterm job

Ngalstyan4 (approved):

var77 (approved):

PR #250, opened 2023-12-18
Add function to reindex indexes created externally

var77:
- Add `lantern_reindex_external_index` function which will get the function pointer from lantern_extras `_lantern_reindex_external_index` function and call it passing the index params got from index header
- Corresponding PR in lantern_extras https://github.com/lanterndata/lantern_extras/pull/49
- Add new tests for lantern_extras function, we will install lantern extras in Linux platforms and run the tests if extension is installed. It will test the external index creation and reindexing
- Throw error on BuildIndex if index file does not exists and say to use `lantern_reindex_external_index` function  #247

### DIFF
--- a/src/hnsw/build.c
+++ b/src/hnsw/build.c
@@ -6,6 +6,7 @@
 #include <assert.h>
 #include <catalog/index.h>
 #include <catalog/namespace.h>
+#include <catalog/pg_proc.h>
 #include <catalog/pg_type.h>
 #include <executor/executor.h>
 #include <fmgr.h>
@@ -18,6 +19,7 @@
 #include <utils/builtins.h>
 #include <utils/lsyscache.h>
 #include <utils/memutils.h>
+#include <utils/syscache.h>
 
 #include "usearch.h"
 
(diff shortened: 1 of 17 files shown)


src/hnsw/build.c
```diff
@@ -531,3 +541,75 @@ void ldb_ambuildunlogged(Relation index)
...
+{
+    BlockNumber          HEADER_BLOCK = 0;
+    Relation             index_rel;
+    Buffer               buf;
+    Page                 page;
+    char                *metric_kind;
+    char                *index_name;
+    HnswIndexHeaderPage *headerp;
+    FmgrInfo             reindex_finfo = {0};
+    char    *function_sig = "_lantern_reindex_external_index(text, text, integer, integer, integer, integer)";
```
Ngalstyan4, line 555: should this function be schema qualified to `lantern`?
Otherwise, a non-root user can just create a function by this name and signature, and have us call it in privileged mode from here

src/hnsw/build.c
```diff
@@ -531,3 +541,75 @@ void ldb_ambuildunlogged(Relation index)
...
+    Oid      function_oid;
+    uint32_t dim = 0;
+    uint32_t m = 0;
+    uint32_t ef_construction = 0;
+    uint32_t ef = 0;
+
+    PG_TRY();
+    {
+        function_oid = DatumGetObjectId(DirectFunctionCall1(regprocedurein, CStringGetDatum(function_sig)));
```
Ngalstyan4, line 564: Is there a reason to prefer the tr/catch approach? 
There are more granular function lookup APIs that will not need exceptions.
(I just have not seen the try/catch machinery be used for this inside postgres codebase, if it is, then probably fine to leave it as is, since this is probably shorter)

test/sql/hnsw_extras.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE EXTENSION lantern_extras;
+-- Validate error on invalid params
+SELECT lantern_create_external_index('v','sift_base1k', 'public', 'invalid_metric',  3, 10, 10, 10);
+SELECT lantern_create_external_index('v','sift_base1k', 'public', 'l2sq',  3, -1, 10, 10);
+SELECT lantern_create_external_index('v','sift_base1k', 'public', 'l2sq',  3, 10, -2, 10);
+SELECT lantern_create_external_index('v','sift_base1k', 'public', 'l2sq',  3, 10, 10, -1);
+\set ON_ERROR_STOP on
+
+-- Create with defaults
+SELECT lantern_create_external_index('v', 'sift_base1k');
```
Ngalstyan4, line 17: Let's add a test for when the underlying table is empty and `dim` is not set

src/hnsw/build.c
```diff
@@ -531,3 +541,75 @@ void ldb_ambuildunlogged(Relation index)
...
+        case usearch_metric_hamming_k:
+            metric_kind = "hamming";
+            break;
+        default:
+            metric_kind = NULL;
+            ldb_invariant(true, "Unsupported metric kind");
+    }
+
+    index_name = pstrdup(RelationGetRelationName(index_rel));
+    dim = headerp->vector_dim;
```
Ngalstyan4, line 596: we should check if `dim` is set, and bail out if it is not (meaning the table is empty and dimension was not explicitly specified)

src/hnsw/build.c
```diff
@@ -531,3 +541,75 @@ void ldb_ambuildunlogged(Relation index)
...
+
+    UnlockReleaseBuffer(buf);
+    relation_close(index_rel, AccessShareLock);
+
+    fmgr_info(function_oid, &reindex_finfo);
+
+    assert(reindex_finfo.fn_addr != NULL);
+
+    DirectFunctionCall6(reindex_finfo.fn_addr,
+                        CStringGetTextDatum(index_name),
```
Ngalstyan4, line 609: Can the other function identify this index by OID in place of a name?
name is not immutable. Since above we take and release a shared lock, it in theory is possible that the name of the index changes between `UnlockReleaseBuffer` buffer and whatever locking the reindexing function does.

The other parameters do not have this issue since those are not changeable.

Ngalstyan4 (approved):
Looks good! you can merge when the rust stuff is ready!

PR #251, opened 2023-12-19
Pick up index dimension on first insert for an empty table

therealdarkknight:
Addresses [issue 198](https://github.com/lanterndata/lantern/issues/198) using the approach documented in the discussion of [PR 209](https://github.com/lanterndata/lantern/pull/209).

When creating an index with no dimension specified on an empty table, we set the dimension in the index header to 0. Later, during the first insert, we infer the dimension from the inserted first row and then update it in the index header.

### DIFF
--- a/test/sql/hnsw_create.sql
+++ b/test/sql/hnsw_create.sql
@@ -36,9 +36,9 @@ CREATE TABLE small_world4 (
     id varchar(3),
     vector real[]
 );
--- If the first row is NULL we do not infer a dimension
+-- If the first inserted row is NULL: we can create an index but we can't infer the dimension from the first inserted row (since it is null)
 \set ON_ERROR_STOP off
-CREATE INDEX ON small_world4 USING hnsw (vector) WITH (M=14, ef=22, ef_construction=2);
+CREATE INDEX first_row_null_idx ON small_world4 USING hnsw (vector) WITH (M=14, ef=22, ef_construction=2);
 begin;
 INSERT INTO small_world4 (id, vector) VALUES
 ('000', NULL),
@@ -46,6 +46,7 @@ INSERT INTO small_world4 (id, vector) VALUES
 CREATE INDEX ON small_world4 USING hnsw (vector) WITH (M=14, ef=22, ef_construction=2);
 rollback;
 \set ON_ERROR_STOP on
+DROP INDEX first_row_null_idx;
 
 INSERT INTO small_world4 (id, vector) VALUES
 ('000', '{1,0,0,0}'),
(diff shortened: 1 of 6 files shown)


var77 (approved):
Looks good.
Do you think adding this code
```c
isNull = true;
while(isNull) {
    // Get the indexed column out of the row and return it's dimensions
    datum = heap_getattr(tuple, indexCol, RelationGetDescr(heap), &isNull);
    if(!isNull) {
        array = DatumGetArrayTypePCopy(datum);
        n_items = ArrayGetNItems(ARR_NDIM(array), ARR_DIMS(array));
        break;
    }
    tuple = heap_getnext(scan, ForwardScanDirection);
    if(tuple == NULL) {
        heap_endscan(scan);
        return 0;
    }
}
```
In line [246](https://github.com/lanterndata/lantern/pull/251/files#diff-8be1a038b91e30250283dd189a9f543a70b16214aa02f9e853646769604c0d97R246) makes sense?

Because now this will fail to infer the dimensions:
```sql
create table test_emp (v real[]);
insert into test_emp (v) values (NULL),( '{1,1,1}');
create index on test_emp using hnsw(v);
```

cc: @Ngalstyan4

therealdarkknight:
> Looks good. Do you think adding this code
> 
> ```c
> isNull = true;
> while(isNull) {
>     // Get the indexed column out of the row and return it's dimensions
>     datum = heap_getattr(tuple, indexCol, RelationGetDescr(heap), &isNull);
>     if(!isNull) {
>         array = DatumGetArrayTypePCopy(datum);
>         n_items = ArrayGetNItems(ARR_NDIM(array), ARR_DIMS(array));
>         break;
>     }
>     tuple = heap_getnext(scan, ForwardScanDirection);
>     if(tuple == NULL) {
>         heap_endscan(scan);
>         return 0;
>     }
> }
> ```
> 
> In line [246](https://github.com/lanterndata/lantern/pull/251/files#diff-8be1a038b91e30250283dd189a9f543a70b16214aa02f9e853646769604c0d97R246) makes sense?
> 
> Because now this will fail to infer the dimensions:
> 
> ```sql
> create table test_emp (v real[]);
> insert into test_emp (v) values (NULL),( '{1,1,1}');
> create index on test_emp using hnsw(v);
> ```
> 
> cc: @Ngalstyan4

Yes, great point! I agree with this change. As far as I know, this behavior was present prior to this PR/issue (I just realized I also brought it up earlier, [see number 2 in my first PR for this ](https://github.com/lanterndata/lantern/pull/209)).

I have some other ideas on how the way we do dimension inference can be refactored-- see number 3 in my first PR above. Since this edit is relevant to index creation on non-empty tables, (and this issue/PR is concerned with index creation on empty tables) I can submit a separate PR with this change and also refactor those functions there. 

What do you think?

var77:
Yep that will work! Actually maybe we won't need that as per discussion with Narek we decided that the tradeoff of UX doesn't worth the complexity added (counting edge cases as well)

test/sql/hnsw_create.sql
```diff
@@ -78,3 +79,30 @@ UPDATE small_world4 SET vector = '{0,0,0}' WHERE id = '001';
...
+    v REAL[]
+);
+
+-- We can still create an index despite having an empty table and not specifying a dimension during index creation
+CREATE INDEX small_world5_hnsw_idx ON small_world5 USING hnsw (v dist_l2sq_ops);
+
+begin;
+-- Our index then infers the dimension from the first inserted row
+INSERT INTO small_world5 (id, v) VALUES
+('000', '{1,0,0,0,1}'),
```
Ngalstyan4, line 95: Can you insert a `NULL` value here first, before inserting anything else, to trigger [this case](https://github.com/lanterndata/lantern/pull/251/files#diff-757fbd016cf447a1d7cb23e47b01cd1d57a1e9f8325084e30d6d9564c7fdc01bR91) ?

Ngalstyan4 (approved):
Small ask, then should be good to go!

PR #252, opened 2023-12-21
mmap index file to memory to reduce memory usage

var77:
Previously we were copying usearch index to provided `result_buf` after the call `usearch_save` we were consuming memory of 2x of index size and then while storing the index to postgres blocks the memory consumption was 3x of index size.
Now we will write index to a file, `mmap` that file into buffer and immediately free the `usearch_index` thus making the memory consumption at most 2x of the index size.
If the index file will be provided externally via `_experimental_index_path` option, we `mmap` that file and skip `usearch_save` step.

### DIFF
--- a/src/hnsw/build.c
+++ b/src/hnsw/build.c
@@ -12,6 +12,9 @@
 #include <miscadmin.h>
 #include <nodes/execnodes.h>
 #include <storage/bufmgr.h>
+#include <sys/fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
 #include <utils/array.h>
 #include <utils/lsyscache.h>
 #include <utils/memutils.h>
@@ -413,6 +416,13 @@ static void BuildIndex(
 {
     usearch_error_t        error = NULL;
     usearch_init_options_t opts;
+    struct stat            index_file_stat;
+    char                  *result_buf = NULL;
+    char                  *index_file_path = NULL;
+    File                   index_file_fd;
+    usearch_metadata_t     metadata;
+    size_t                 num_added_vectors;
+
     MemSet(&opts, 0, sizeof(opts));
 
     InitBuildState(buildstate, heap, index, indexInfo);
@@ -434,7 +444,7 @@ static void BuildIndex(
         }
         elog(INFO, "done loading usearch index");
 
-        usearch_metadata_t metadata = usearch_metadata(buildstate->usearch_index, &error);
+        metadata = usearch_metadata(buildstate->usearch_index, &error);
         assert(error == NULL);
         opts.connectivity = metadata.connectivity;
         opts.dimensions = metadata.dimensions;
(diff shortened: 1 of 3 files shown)


src/hnsw/build.c
```diff
@@ -482,25 +492,50 @@ static void BuildIndex(
...
-    usearch_save(buildstate->usearch_index, NULL, &result_buf, &error);
-    assert(error == NULL && result_buf != NULL);
-
-    size_t num_added_vectors = usearch_size(buildstate->usearch_index, &error);
+    metadata = usearch_metadata(buildstate->usearch_index, &error);
     assert(error == NULL);
 
+    if(buildstate->index_file_path == NULL) {
+        // size of static name + max digits of uint32 (Oid) + 1 for nullbyte
+        index_file_path = palloc0(sizeof("/tmp/ldb-index-.bin") + 10 + 1);
```
Ngalstyan4, line 500: * Factor out the calculation and the format string constant to a local variable
* use the constant format string in the sprintf call, so if we change the file path, we will be sure both allocation and formatting will be updated
* use snprintf or other sized string formatter to make sure we do not silently buffer-overflow
* make sure the size parameter passed to snprintf is a compile time constant (it is in this case as it is sizeof() + constant)

src/hnsw/build.c
```diff
@@ -482,25 +492,50 @@ static void BuildIndex(
...
     buildstate->usearch_index = NULL;
+    //****************************** mmap index to memory END ******************************//
+
+    //****************************** saving to WAL BEGIN ******************************//
+    UpdateProgress(PROGRESS_CREATEIDX_PHASE, PROGRESS_HNSW_PHASE_LOAD);
+    StoreExternalIndex(index, &metadata, forkNum, result_buf, &opts, num_added_vectors);
+
+    //****************************** saving to WAL END ******************************//
+
+    munmap(result_buf, index_file_stat.st_size);
```
Ngalstyan4, line 531: this can also `MAP_FAILED` right?

src/hnsw/build.c
```diff
@@ -482,25 +492,50 @@ static void BuildIndex(
...
+    close(index_file_fd);
+
+    if(buildstate->index_file_path == NULL) {
+        // remove index file if it was not externally provided
+        unlink(index_file_path);
```
Ngalstyan4, line 536: can you store the tmp index_file_path in another var called `tmp_inde_file_path`, and make this branch conditional on `sprintf`ed `tmp_inde_file_path` existing?
The way it stands, it is confusing and I am afraid some refactoring may end us up deleting users' externally managed files

src/hnsw/build.c
```diff
@@ -482,25 +495,51 @@ static void BuildIndex(
...
+    metadata = usearch_metadata(buildstate->usearch_index, &error);
     assert(error == NULL);
 
+    if(buildstate->index_file_path == NULL) {
+        // Save index into temporary file
+        // To later mmap it into memory
+        // Filename is /tmp/ldb-index-$relfilenode.bin
+        // The file will be removed in the end
+        tmp_index_file_path = palloc0(tmp_index_file_char_cnt);
+        sprintf(tmp_index_file_path, tmp_index_file_fmt_str, index->rd_rel->relfilenode);
```
Ngalstyan4, line 507: let's use `snprintf` to make sure we do not overflow

Ngalstyan4: or do an assert on the returned value of `sprintf`

src/hnsw/build.c
```diff
@@ -482,25 +495,51 @@ static void BuildIndex(
...
-    StoreExternalIndex(index, buildstate->usearch_index, forkNum, result_buf, &opts, num_added_vectors);
-
+    StoreExternalIndex(index, &metadata, forkNum, result_buf, &opts, num_added_vectors);
     //****************************** saving to WAL END ******************************//
 
-    usearch_free(buildstate->usearch_index, &error);
-    free(result_buf);
-    assert(error == NULL);
-    buildstate->usearch_index = NULL;
+    assert(munmap(result_buf, index_file_stat.st_size) == 0);
```
Ngalstyan4, line 535: in `release` builds `assert`s are compiled away (it is a macro, not a function, so everything in the parentheses gets dropped).
So, the munmap call has to be outside of `assert`

Issue #125, opened 2023-09-08
Add support for unlogged tables


Ngalstyan4:
Currently, in the code we hardcode `MAIN_FORK` as fork number.
But unlogged tables (tables belonging to the current session and not persisted in WAL or elsewhere) live on a separate fork (INIT_FORK I think?)

We need to change the hardcoded MAIN_FORKs in the code and make sure all LanternDB functionality works smoothly on unlogged tables

PR #253, opened 2023-12-21
Unlogged tables support

therealdarkknight:
Resolves [issue 125](https://github.com/lanterndata/lantern/issues/125).

It makes sure that preexisting functionality for unlogged table index creation and insertion of rows works, and also properly implements the `ambuildempty` access method function by constructing an empty index and storing it in the init fork.

Two new tests have been added: `hnsw_insert_unlogged` and `hnsw_create_unlogged`. They are the unlogged versions of the `hnsw_insert` and `hnsw_create` tests respectively.

Also, a new directory `manual_tests` have been added. The two files in here are to test the functionality of these unlogged tables and indexes after a postgres crash. Instructions are in these test files, but they are meant to be run consecutively with a `kill -9` of postgres in between (hence the "manual"). A possible todo is to automate this in a script and add it to our testing infra. I ran these manual tests prior to pushing changes.


var77 (approved):
Looks good!
Can we also add tests when the table will be changed from unlogged to logged an vice-versa? 

Also we may add tests in replication mode if you that it's worth that.

One test case that can be done I think in replica mode is:
1. Create index on unlogged table
2. Make table logged
3. Insert data on primary node
4. Crash and restart slave and call validate_index on it.

Similar test example can be found [here](https://github.com/lanterndata/lantern/blob/main/test/c/replica_test_index.c)

If you decide to write a new test with replica you will need to name the file like `replica_test_$testname` and then include it in this two places [[1]](https://github.com/lanterndata/lantern/blob/main/test/c/replica_test_index.c) [[2]](https://github.com/lanterndata/lantern/blob/main/test/c/runner.c#L101)

var77:
Just noticed the sanitizer tests are crashing the database is it connected to the PR? You can check the error message [here](https://github.com/lanterndata/lantern/actions/runs/7293284024/job/19876020289?pr=253#step:14:850)

therealdarkknight:
Hey! I fixed a bug that was causing the sanitizer tests to crash.

I can work on the replication tests now. Is there an easy way to run the replication tests locally?

I'm looking through the ci scripts and trying to get the replicate environment running locally but I'm running into various environment-related issues. Can keep debugging it but wanted to see if there's something available already

var77:
Hi! Sorry forgot to mention how to do replication tests locally.

```bash
docker build -f Dockerfile.dev --build-arg VERSION=15 . -t lantern-pg-debug-15
docker run --cap-add=SYS_PTRACE -p 4444:5432 --name lantern-15 -ti --rm lantern-pg-debug-15 bash

# Then inside container
. ci/scripts/bitnami-utils.sh
start_postgres_master
start_postgres_replica

# Then run tests
cd build
sudo PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/pgsql/bin ENABLE_REPLICA=1 REPLICA_PORT=5443 DB_PORT=5442 make test-client
```

Let me know if you encounter any issues

var77 (approved):

therealdarkknight:
There was a bug in `runner.c` where the `replica_test_index` test had to be the last one to run because it crashed the root db as part of its test. I fixed it in the most recent commit by reconnecting the root db connection on every test case (and increasing the sleep duration to let the replica sync more reliably). This way, we can write multiple tests that crash the root db and run them consecutively

It looks like one of the tests is failing because it could not reconnect the root db for another test case. Could this be because it's the first time we're running multiple replicate tests-- anything that needs to change in the workflow to accommodate that?

var77:
Can you try to add these 2 changes, I hope that should help, I guess the postgres backend is being killed from another test running in parallel.

https://github.com/var77/lantern/pull/7/commits/ab9c05754a3fd45d8653e58e375e9f7857469c09

-- This one is for runner but you should also delete the last line (:196) calling `PQfinish(root_conn);` or it will throw double free issue

https://github.com/var77/lantern/pull/7/commits/4df904b4f055b5d9396b4710f9e9b81bc006037a

therealdarkknight:
Looks like it's resolved the original issue but now we run the risk of not having a postgres instance for one of the update tests?

var77:
> Looks like it's resolved the original issue but now we run the risk of not having a postgres instance for one of the update tests?

Hm I am not sure actually why this would happen in the first place I was thinking that the runners should be isolated, but maybe we can assign a random free port or port number based on the postgres version? And also get the PID like this 
```bash
/postgres &
POSTGRES_PID=$!
# do other stuff
kill -9 $POSTGRES_PID
```
instead of using `fuser` ?

PR #257, opened 2024-01-14
Stop postgres using pg_ctl to avoid race conditions

var77:
There were 2 issues:
1. `pg_isready` was getting permissions issue when running from /home/runner/work/lantern/lantern (`pg_response could not change directory to "/home/runner/work/lantern/lantern": Permission denied`) . So I am changing directory to `/tmp` before running it.
2. When just killing postgres process via PID it needs some time before completely shut down. So if we kill postgres and immediately run again, the run might fail because some filesystem locks. Now we will stop postgres using `pg_ctl` which will wait for the process to completely stop before returning

### DIFF
--- a/ci/scripts/bitnami-utils.sh
+++ b/ci/scripts/bitnami-utils.sh
@@ -856,6 +856,9 @@ crash_and_restart_postgres() {
   if [[ "$pid" != "" ]]; then
     kill -9 $pid
   fi
-
+  
+  # Be sure that postgres is stopped before starting again
+  pg_ctl stop -D $POSTGRESQL_DATA_DIR || true
+  
   postgresql_start_bg
 }
--- a/ci/scripts/run-tests-linux.sh
+++ b/ci/scripts/run-tests-linux.sh
@@ -10,7 +10,7 @@ export PGDATA=/etc/postgresql/$PG_VERSION/main
 
 function stop_current_postgres() {
   # Stop any existing processes
-  /usr/lib/postgresql/$PG_VERSION/bin/pg_ctl stop -D $PGDATA || true
+  /usr/lib/postgresql/$PG_VERSION/bin/pg_ctl stop -D $PGDATA
 }
 function wait_for_pg(){
  tries=0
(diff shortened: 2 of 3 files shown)


ci/scripts/bitnami-utils.sh
```diff
@@ -856,6 +856,9 @@ crash_and_restart_postgres() {
   if [[ "$pid" != "" ]]; then
     kill -9 $pid
   fi
-
+  
+  # Be sure that postgres is stopped before starting again
+  pg_ctl stop -D $POSTGRESQL_DATA_DIR || true
```
Ngalstyan4, line 861: does this ever fail? is `||true` necessary?

ci/scripts/run-tests-linux.sh
```diff
@@ -8,6 +8,10 @@ RUN_TESTS=${RUN_TESTS:-1}
 
 export PGDATA=/etc/postgresql/$PG_VERSION/main
 
+function stop_current_postgres() {
+  # Stop any existing processes
+  /usr/lib/postgresql/$PG_VERSION/bin/pg_ctl stop -D $PGDATA || true
```
Ngalstyan4, line 13: does this ever fail? is `||true` necessary?

Ngalstyan4 (approved):
Looks good!
I am worried about these `||true`s throughout our shell scripts.
Do we expect the commands to fail that is why we are masking the error values?
If so, in what cases does it fail and why can we still continue running the script?

var77:
I think for now it should not fail, I will try to remove || true's and see. It will fail if there won't be postgres process running but we should not have that case

ci/scripts/bitnami-utils.sh
var77: I will leave this one, as this function will be used in `crash_and_restart_postgres_master` and `crash_and_restart_postgres_replica` functions, which are killing process with `kill -9` signal, so after that I am stopping postgres with `pg_ctl` just in case to be sure that postgres is 100% stopped before starting it again. If it won't be fully stopped we will get error on start and the incoming test case will fail to connect to db

PR #260, opened 2024-01-18
Add test to catch external index with missing op class

var77:
When we were creating an external index we were not passing the operator class, letting the index to read it from file. This was okay for index creation, but when doing queries for example with <=> (cosine) operator the postgres was not detecting an index scan as <=> operator is connected with dist_cos_ops operator class and if we don't specify the operator class it will fallback to l2sq_dist_ops by default.

There's also an issue with generic `<?>` operator, the `EXPLAIN` query crashes the database, I have currently commented that statement, so other tests won't get affected by it.


ci/scripts/build-linux.sh
```diff
@@ -27,7 +27,7 @@ function install_platform_specific_dependencies() {
   # Currently lantern_extras binaries are only available for Linux x86_64
   # We won't install onnxruntime as lantern_extras are used only for external index in tests
   pushd /tmp
-    LANTERN_EXTRAS_VERSION=0.0.6
+    LANTERN_EXTRAS_VERSION=0.0.12
```
Ngalstyan4, line 30: Are there any downsides to pulling the latest version of extras via github API ([https://api.github.com/repos/lanterndata/lantern_extras/releases/latest](https://api.github.com/repos/lanterndata/lantern_extras/releases/latest)) and using that in the build here?


[1]: https://api.github.com/repos/lanterndata/lantern_extras/releases/latest

test/sql/hnsw_extras.sql
```diff
@@ -20,12 +20,33 @@ SELECT lantern_create_external_index('v', 'empty');
...
 SELECT lantern_create_external_index('v', 'sift_base1k');
 SELECT _lantern_internal.validate_index('sift_base1k_v_idx', false);
+
+SELECT v AS v777 FROM sift_base1k WHERE id = 777 \gset
+-- Validate that using corresponding operator triggers index scan
+SET lantern.pgvector_compat=TRUE;
+EXPLAIN (COSTS FALSE) SELECT id FROM sift_base1k order by v <-> :'v777' LIMIT 10;
+
+-- =========== TODO:: THIS CAUSES SERVER CRASH =============== -
+-- SET lantern.pgvector_compat=FALSE;
```
Ngalstyan4, line 30: Let's move this to hnsw_todos

PR #262, opened 2024-01-22
Narek/storage upgrade

Ngalstyan4:


src/hnsw/validate_index.c
```diff
@@ -294,28 +293,15 @@ static void ldb_vi_read_node_carefully(void               *node_tape,
...
     unsigned tape_pos = 0;
-    uint32   level_on_tape;
+    uint16   level_on_tape;
     uint32   neighbors_nr;
     uint32   neighbors_max;
     uint32  *neighbors;
     uint32   unused;
 
     LDB_VI_READ_NODE_CHUNK(vi_node, vi_node->vn_label, node_tape, &tape_pos, node_tape_size);
-    LDB_VI_READ_NODE_CHUNK(vi_node, vi_node->vn_dim, node_tape, &tape_pos, node_tape_size);
```
Ngalstyan4, line 304: deleted because new binary index does not contain this information, so there is no room for inconsistency

PR #265, opened 2024-01-23
Narek/release0.0.12

Ngalstyan4:


CMakeLists.txt
```diff
@@ -311,7 +333,7 @@ if (CLANG_FORMAT)
 
   string(REGEX MATCH "[0-9]+\\.[0-9]+\\.[0-9]+" CLANG_FORMAT_VERSION "${CLANG_FORMAT_VERSION}")
 
-  if(CLANG_FORMAT_VERSION VERSION_LESS 14)
+  if(CLANG_FORMAT_VERSION VERSION_LESS 13)
     message(WARNING "clang-format version ${CLANG_FORMAT_VERSION} found, need at least 14")
```
var77, line 337: Should we change the version number in error message as well?

src/hnsw/utils.c
```diff
@@ -177,6 +177,17 @@ bool VersionsMatch()
         version_checked = true;
 
         SPI_finish();
+
+        if(!versions_match) {
+            elog(WARNING,
```
var77, line 182: Do we need the elog to be moved here? As now we will have duplicate logs for build and insert (as they're changed to WARNING) right?

var77 (approved):
Looks good!

src/hnsw/utils.c
Ngalstyan4: this is logged once, as otherwise `versions_match` is cached.
it is good to have here for debugging, since it actually prints the version names and tells something about what the mismatch is

PR #276, opened 2024-02-01
Fixes for bugs in cost estimate function (relevant for partial indexes)

therealdarkknight:
1. There was an unsigned integer underflow bug in our cost estimation that would lead to astronomically high costs for certain partial indexes. To replicate (check out to first commit), delete the `/tmp/lantern/vector_datasets` folder (to force download of a new dataset file) and `make test`, and look at the output of the `hnsw_cost_estimate` test, at the end. The cost value for the added test case is extremely high, indicating an unsigned underflow:

```
+DEBUG:  LANTERN - Query cost estimator
+DEBUG:  LANTERN - ---------------------
+DEBUG:  LANTERN - Total cost: 25261826802294276096.000000
+DEBUG:  LANTERN - Selectivity: 1.000000
+DEBUG:  LANTERN - Num pages: 6148299799767393280.000000
+DEBUG:  LANTERN - Num tuples: 58.000000
+DEBUG:  LANTERN - ---------------------
``` 

This underflow prevents certain partial indexes from being used-- which we can see in this test as well: there is a lantern partial index, and a filtered query whose filter matches exactly that of the partial index, and so the partial index should be used but it is not (because of its extremely high cost). 

We fix this underflow (check out to second commit) by casting to signed int in the appropriate calculation.

2. We were previously using `path->indexinfo->rel->tuples` as the number of tuples in our index, but this is actually the number of tuples in the table. This is problematic when we deal with partial indexes, when the number of tuples in the index can differ significantly from the number of tuples in the table. The number of tuples in our index is given by `path->indexinfo->tuples`.

3. There is another potential unsigned integer underflow in `estimate_number_tuples_accessed` when we calculate this line:
`uint64 total_tuple_visits = tuples_visited_per_non_base_level * (expected_num_levels - 1);`. If we check that `num_tuples_in_index > 0` earlier in the function, it follows that `expected_num_levels` is always at least 1, and so we won't underflow.

### DIFF
--- a/scripts/run_all_tests.sh
+++ b/scripts/run_all_tests.sh
@@ -46,6 +46,7 @@ then
         curl -sSo sift_base1k.csv https://storage.googleapis.com/lanterndb/sift_base1k.csv
         curl -sSo siftsmall_base.csv https://storage.googleapis.com/lanterndata/siftsmall/siftsmall_base.csv
         curl -sSo tsv_wiki_sample.csv https://storage.googleapis.com/lanterndb/tsv_wiki_sample.csv
+        curl -sSo views_vec10k.csv https://storage.googleapis.com/lanterndata/random_multicolumn/views_vec10k.csv
         # Convert vector to arrays to be used with real[] type
         cat sift_base1k.csv | sed -e 's/\[/{/g' | sed -e 's/\]/}/g' > sift_base1k_arrays.csv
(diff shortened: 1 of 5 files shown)


Ngalstyan4:
Good catch and fix @therealdarkknight ! thanks!

PR #277, opened 2024-02-01
Add product quantization

var77:
Added Product Quantization feature.

Overview:
1. Provide SQL function `quantize_table(tbl regclass, col NAME, cluster_cnt INT,subset_count INT, distance_metric TEXT)`
2. The function will call `create_codebook(tbl REGCLASS, col NAME, cluster_cnt INT, subset_count INT, distance_metric TEXT)`
3. Create codebook will scan the whole table and run `k-means` over dataset for the provided vector subsets and cluster count
4. Then this will return 3 dimensional array which will be inserted into the codebook table `_lantern_codebook_${table_name}` (subvector_id, centroid_id)
5. Then vector compression will be applied on the table and new column will be created of type `PQVector`. This will help us to save storage compared to regular `INT[]`.
6. Then you can do queries over the quantized vectors using `decompress_vector(v pqvec, codebook regclass)` function

TODO
- [x] Use generated columns for postgres > 11 instead of triggers


src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+        for(i = 0; i < *used_indexes_size; i++) {
+            if(used_indexes[ i ] == idx) {
+                used = true;
+                break;
+            }
+        }
+
+        if(!used) break;
+    }
+    used_indexes[ *used_indexes_size ] = idx;
```
Ngalstyan4, line 36: who sizes `used_indexes` to make sure that this does not go out of bounds?

It seems you only want to write to `used_indexes` and `used_indexes_size` to append a single value to the end of the array that is the same value as the returned `idx` right?
if so, might be cleaner to make `used_indexes` and `used_indexes_size` `const` within this function, and update the arrays at callsite

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+    return idx;
+}
+
+/*
+ * Initialize k clusters by chosing random points from the dataset
+ */
+static Cluster **initialize_clusters(uint32 k, float4 **dataset, uint32 dataset_size, uint32 subset_start)
+{
+    uint32    i;
+    Cluster **clusters = palloc(sizeof(Cluster) * k);
```
Ngalstyan4, line 47: shouldn't this be `sizeof(Cluster*)` ?

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+    return distance <= threshold;
+}
+
+/*
+ * Run k-means clustering over the dataset in a subset of vectors
+ * args:
+ *   k - cluster count
+ *   dataset - 2d array of float numbers
+ *   dataset_size - length of dataset (axis 0)
+ *   subset_start - index from which the subset will be taken (in case if subset count is 1 this will be 0)
```
Ngalstyan4, line 188: > in case if subset count is 1 this will be 0

what does this mean?

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+                 uint32                dataset_size,
+                 uint32                subset_start,
+                 uint32                subset_dim,
+                 usearch_metric_kind_t distance_metric,
+                 uint32                iter)
+{
+    uint32    i, j;
+    float4  **centroids = palloc(sizeof(size_t) * k);
+    Cluster **clusters = initialize_clusters(k, dataset, dataset_size, subset_start);
+    float4  **old_centers = palloc(sizeof(size_t) * k);
```
Ngalstyan4, line 205: are `old_centers` the values of `centroids` from previous iteration?

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+ * */
+static float4 *calculate_mean(float4 **points, uint32 point_cnt, uint32 subset_dim)
+{
+    float4 *mean = palloc0(sizeof(float4) * subset_dim);
+    uint32  i;  // rows
+    uint32  j;  // cols
+
+    for(i = 0; i < point_cnt; i++) {
+        for(j = 0; j < subset_dim; j++) {
+            if(!mean[ j ]) mean[ j ] = 0;
```
Ngalstyan4, line 123: does this do anything?

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+        centroids[ i ] = clusters[ i ]->center;
+    }
+    return centroids;
+}
+
+/*
+ * Run k_means over each subset of dataset vectors
+ * args:
+ *   cluster_count - number of clusters to initialize for each subset. Then each subset will have centroids equal to
+ *   subset_count - how many parts split the vectors of dataset. Each subset will be trained separately and
```
Ngalstyan4, line 231: Were the two sentences truncated in the end?

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+ *   dataset - 2D array of f32 values
```
Ngalstyan4, line 232: where each row is a vector that we are PQ-quantizing, right?

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+                                uint32                dataset_size,
+                                uint32                dim,
+                                usearch_metric_kind_t distance_metric,
+                                uint32                iter)
+{
+    uint32     i;
+    uint32     subset_start;
+    uint32     subset_dim;
+    float4   **subset_centroids;
+    Codebook  *current_codebook;
```
Ngalstyan4, line 252: call the type PQCodebook

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+    Codebook **codebooks = palloc(sizeof(Codebook) * subset_count);
```
Ngalstyan4, line 253: `sizeof(Codebook*)`?

src/hnsw/product_quantization.h
```diff
@@ -0,0 +1,31 @@
+#ifndef LDB_HNSW_PRODUCT_QUANTIZATION_H
+#define LDB_HNSW_PRODUCT_QUANTIZATION_H
+#include <postgres.h>
+
+#include <utils/builtins.h>
+
+#include "usearch.h"
+
+typedef struct Codebook
```
Ngalstyan4, line 9: no need for `Codebook` on this line

test/sql/hnsw_pq.sql
```diff
@@ -0,0 +1,75 @@
...
+\ir utils/random_array.sql
+\ir utils/calculate_recall.sql
+
+SET client_min_messages=ERROR;
+\set ON_ERROR_STOP off
+-- Test PQVec type
+SELECT '{84,1,4,128,256}'::pqvec;
+SELECT '{84,1,4,128,255}'::pqvec;
+SELECT '{84,1,4,128,255}'::pqvec::INT[];
+SELECT '{84,1,4,128,255}'::INT[]::pqvec;
```
Ngalstyan4, line 12: These should not error right?
should they be above line 7?

test/sql/hnsw_pq.sql
```diff
@@ -0,0 +1,75 @@
...
+SELECT l2sq_dist(:'decompressed_1', :'decompressed_2');
+
+-- Test recall for quantized vs non quantized vectors
+ALTER TABLE sift_base1k ADD COLUMN v_pq_dec REAL[];
+UPDATE sift_base1k SET v_pq_dec=decompress_vector(v_pq, '_lantern_codebook_sift_base1k');
+-- Calculate recall over original vector
+SELECT (calculate_table_recall('sift_base1k', 'sift_query1k', 'sift_truth1k', 'v', 10, 100) -
+       calculate_table_recall('sift_base1k', 'sift_query1k', 'sift_truth1k', 'v_pq_dec', 10, 100)) as recall_diff \gset
+
+SELECT :'recall_diff' < 0.2 as recall_diff_meets_threshold;
```
Ngalstyan4, line 63: I think once we use postgres-controlled `rand()` in clustering, we should be able to `setseed()` and output actual recall values instead of this

Ngalstyan4 (commented):
Great work Varik jan!

And, happy birthday 🎉

Some cases to test
- Have a table on a non-default schema and generate codebook on it
- Pass a non-existent distance metric to compress_vector


It would be nice to do dependency tracking between the main table and the pq_codebook table so when one deletes the vector column on which the codebook was created, the codebook gets deleted.
We can probably achieve this with triggers. Let's not do it now, though, unless it is trivial

var77:
Thank you! 

Thanks for detailed review will address the comments tmrw/mon and we'll be good to merge this!

sql/lantern.sql
```diff
@@ -193,3 +195,223 @@ BEGIN
...
+      EXECUTE stmt;
+  	END LOOP;
+  END LOOP;
+
+  return codebook_table;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Compress vector using codebook
+CREATE OR REPLACE FUNCTION _lantern_internal.compress_vector(v REAL[], subset_count INTEGER, codebook regclass, distance_metric TEXT)
```
Ngalstyan4, line 268: does this need to take `subset_count`?
can't it infer the value cheaply from the codebook?

sql/lantern.sql
```diff
@@ -193,3 +195,223 @@ BEGIN
...
+RETURNS REAL[] AS $$
+DECLARE
+  res REAL[];
+  subset REAL[];
+  centroid_id INT;
+  subvector_id INT;
+BEGIN
+  res := '{}'::REAL[];
+  subvector_id := 0;
+  FOREACH centroid_id in array v::INT[]
```
Ngalstyan4, line 323: This is a public function, so worth doing more thorough argument validation and throw helpful error messages early
e.g. codebook may be a relation that is not a codebook, `v` might be empty or have the wrong size

if `v` is empty, currently this function returns the cryptic `ERROR: FOREACH expression must not be null`, which is not ideal, as the caller has no idea we use FOREACH in here

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+ * This will be used to pick random centroids from the dataset
+ * When initializing clusters
+ * */
+static int get_available_index(int max_idx, int *used_indexes, int *used_indexes_size)
+{
+    int  idx;
+    int  i;
+    bool used = false;
+    while(true) {
+        idx = rand() % (max_idx + 1);
```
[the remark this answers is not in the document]
Ngalstyan4: actually, nevermind about setseed - `setseed` only controls SQL-level calls to `random`, and using that seed from inside our extension may result in undesired behaviour.

just switch to pg_prng as done [here](https://github.com/lanterndata/lantern/blob/main/src/hnsw/insert.c#L36-L40) but ignore the "big reason" part

src/hnsw/product_quantization.c
```diff
@@ -0,0 +1,275 @@
...
+                 uint32                subset_dim,
+                 usearch_metric_kind_t distance_metric,
+                 uint32                iter)
+{
+    uint32    i, j;
+    float4  **centroids = palloc(sizeof(size_t) * k);
+    Cluster **clusters = initialize_clusters(k, dataset, dataset_size, subset_start);
+    float4  **old_centers = palloc(sizeof(size_t) * k);
+
+    for(i = 0; i < iter; i++) {
```
Ngalstyan4, line 207: if it is safe to do so, once in every iteration call CHECK_FOR_INTERRUPTS() to make sure the client is able to cancel the request if they change their mind by sending a signal to the process (e.g. via CTRL-C)

sql/lantern.sql
```diff
@@ -193,3 +195,223 @@ BEGIN
...
+	WITH FUNCTION ldb_cast_pqvec_array(pqvec, integer, boolean) AS ASSIGNMENT;
+	
+CREATE OR REPLACE FUNCTION create_codebook(tbl REGCLASS, col NAME, cluster_cnt INT, subset_count INT, distance_metric TEXT)
+RETURNS NAME AS $$
+DECLARE
+  stmt TEXT;
+  res REAL[];
+  codebooks REAL[][][];
+  i INT;
+  subset_len INT;
```
Ngalstyan4, line 231: unused?

sql/lantern.sql
var77: This one is internal function, theres external wrapper which does infer the value from codebook

src/hnsw/product_quantization.c
var77: If let's say we are quantizing [f32x128] -> [u8x1] so the only range will be [0-128] (subvector_start: 0, subvector_dim: 128), but I will rephrase this comment as for the first subvector it will also be 0

src/hnsw/product_quantization.c
var77: yes

sql/lantern.sql
```diff
@@ -193,3 +195,223 @@ BEGIN
...
+
+-------------------------------------
+-------- Product Quantization -------
+-------------------------------------
+
+CREATE FUNCTION ldb_pqvec_in(cstring, oid, integer) RETURNS pqvec AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+CREATE FUNCTION ldb_pqvec_out(pqvec) RETURNS cstring AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+CREATE FUNCTION ldb_pqvec_recv(internal, oid, integer) RETURNS pqvec AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+CREATE FUNCTION ldb_pqvec_send(pqvec) RETURNS bytea AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+CREATE FUNCTION ldb_cast_array_pqvec(int[], integer, boolean) RETURNS pqvec	AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
```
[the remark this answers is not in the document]
var77: This is not compressing the vector just converting int32 array to byte array. So something like '{1,2,3}' will be converted to byte array and save storage.

src/hnsw.c
```diff
@@ -443,6 +448,153 @@ Datum       lantern_reindex_external_index(PG_FUNCTION_ARGS)
...
+
+    if(dataset_size < cluster_cnt) {
+        elog(ERROR, "Dataset size can not be smaller than cluster count");
+    }
+
+    if(dataset_dim < subset_cnt) {
+        elog(ERROR, "Dataset dimension can not be smaller than subset count");
+    }
+
+    // TODO:: this check can be removed as soon as we resolve return type issue from this function
```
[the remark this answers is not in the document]
var77: I can not create 3d array using `construct_md_array` if inner arrays does not have same dimensions.

PR #279, opened 2024-02-02
Use proper index selectivity in hnswcostestimate function

therealdarkknight:
As per [postgres documentation](https://www.postgresql.org/docs/current/index-cost-estimation.html), `indexSelectivity` is the estimated fraction of parent table rows that will be retrieved during the index scan. This was previously hardcoded to `1`, but it should instead be the fraction of parent table rows that are in our index (which can be less than `1` when we have partial indexes).
The `genericcostestimate` function that we call appears to compute this value for us, and so we can use that instead.

I also added tests in `hnsw_cost_estimate` that compute the `indexSelectivity` values (shown in the lantern debug logs that are outputted during this test) for partial indexes, to make sure that `indexSelectivity` is computed properly.

### DIFF
--- a/src/hnsw/utils.h
+++ b/src/hnsw/utils.h
@@ -41,12 +41,4 @@ static inline void ldb_invariant(bool condition, const char *msg, ...)
         }                              \
     } while(0)
 
-// same as ldb_dlog above but intended for output in tests that needs to be consistent across systems and OS
-#define ldb_tlog(...)                  \
-    do {                               \
-        if(unlikely(ldb_is_test)) {    \
-            elog(DEBUG4, __VA_ARGS__); \
-        }                              \
-    } while(0)
-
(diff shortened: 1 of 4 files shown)


src/hnsw/utils.h
```diff
@@ -41,4 +41,12 @@ static inline void ldb_invariant(bool condition, const char *msg, ...)
         }                              \
     } while(0)
 
+// same as ldb_dlog above but intended for output in tests that needs to be consistent across systems and OS
+#define ldb_tlog(...)                  \
+    do {                               \
+        if(unlikely(ldb_is_test)) {    \
+            elog(DEBUG4, __VA_ARGS__); \
```
Ngalstyan4, line 48: this overrides the definition above. you should just change the log level above and not redefine the macro function

PR #284, opened 2024-02-08
Narek/hnsw scan cache

Ngalstyan4:


Ngalstyan4:
benchmarked and did not see any performance improvement from such caching, so closing.

PR #286, opened 2024-02-17
Narek/pgvector compat

Ngalstyan4:


var77 (approved):
Should we also change `reindex_lantern_indexes` function to only reindex lantern_hnsw indexes? Currently it searches for indexes like %hnsw%

var77:
Also I was thinking if should we somehow mark minimum compatible version of lantern extras and throw error in external indexing functions?

PR #294, opened 2024-02-22
Adding deploy with stitch badge

rdosanjh:
Adds a deploy with Stitch badge to the README.
image

Hey I'm Raj from Stitch (YC W24) and we are building installers for opensource software. We feel like Lantern could really make use of this.

Here is a short video demo:

https://www.loom.com/share/9e205e26abd0467e97de861b50ba20be

The Stitch installer essentially creates infra on the user's cloud and runs a script to install Lantern. We have written the install script using the instructions on the readme. We will be fully opensourcing the Stitch installer so that it is transparent to users exactly what is being run on their cloud.

Please let me know if you have any questions.


Ngalstyan4:
resolved this offline.

PR #298, opened 2024-03-18
Call external index function and import created index file on BuildIndex if external=true

var77:
- Added new option to index `bool external`
- If external is true we will call `_create_external_index` function from lantern_extras
- The multicore index creation will be done in Rust side, index will be saved in file and file path will be returned from the function
- We will then use the file path as `_experimental_index_path` option in our `BuildIndex` function to import that file


var77:
Covered by https://github.com/lanterndata/lantern/pull/324

PR #302, opened 2024-03-30
Add async tasks to lantern

Ngalstyan4:


var77:
I think we should run [this command](https://github.com/lanterndata/lantern/blob/main/ci/scripts/run-tests-linux.sh#L58) conditionally based on `ENABLE_COVERAGE` to avoid the error

var77 (approved):

PR #304, opened 2024-03-30
[doc]: add prerequisites in README.md

### LINKED ISSUE #303: [doc]: README.md does not specify prerequisites for build Lantern from source code
We have prerequisites specified in https://lantern.dev/docs/lantern-db/install, but not in README.md.

broccoliSpicy:
This PR tries to fix #303 by adding prerequisites for building Lantern from source code in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,19 +18,7 @@ If you don’t have PostgreSQL already, use Lantern with [Docker](https://hub.do
 ```bash
 docker run --pull=always --rm -p 5432:5432 -e "POSTGRES_USER=$USER" -e "POSTGRES_PASSWORD=postgres" -v ./lantern_data:/var/lib/postgresql/data lanterndata/lantern:latest-pg15
 ```
-
 Then, you can connect to the database via `postgresql://$USER:postgres@localhost/postgres`.
-
-To install Lantern from source on top of your existing PostgreSQL:
-
-```
-git clone --recursive https://github.com/lanterndata/lantern.git
-cd lantern
-mkdir build
-cd build
-cmake ..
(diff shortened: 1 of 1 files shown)


broccoliSpicy:
after reading 
`
Once a variable is in the cache, its “cache” value cannot normally be modified from a CMakeLists file. The reasoning behind this is that once CMake has put the variable into the cache with its initial value, the user may then modify that value from the GUI. If the next invocation of CMake overwrote their change back to the [set](https://cmake.org/cmake/help/latest/command/set.html#command:set) value, the user would never be able to make a change that CMake wouldn’t overwrite. A set(FOO ON CACHE BOOL "doc") command will typically only do something when the cache doesn’t have the variable in it. Once the variable is in the cache, that command will have no effect.
`
from [cmake cache doc](https://cmake.org/cmake/help/book/mastering-cmake/chapter/CMake%20Cache.html), I think it might be better to specify `MARCH_NATIVE=ON` explicitly.
like this:
```
git clone --recursive https://github.com/lanterndata/lantern.git
cd lantern
mkdir build
cd build
cmake -DMARCH_NATIVE=ON ..
make install
```

I am just worried that the user might build portable binaries first then follow the instructions to build a CPU-specific vectorization one, see that the only difference is `cmake ..` and only reruns this command, without knowing that `MARCH_NATIVE=ON` is actually not taking effect, since the settings we have in `CMakeLists.txt` won't overwrite CMakeCache.txt

Ngalstyan4:
thanks for your contribution, @broccoliSpicy !

PR #324, opened 2024-07-16
External indexing via tcp socket

var77:
add `external` param to index creation
add `lantern_hnsw.external_index_host` GUC.
add `lantern_hnsw.external_index_port` GUC.

When the `external` parameter will be passed we connect to remote lantern daemon  via tcp socket using the `external_index_host` and `external_index_port` GUC variables. First we send the index header to socket and codebook if needed. After getting ack response we start streaming tuples via socket to daemon which will create multithreaded usearch index. Then it will send back the `num_tuples_added` and `index_size` (which is the file size for index) and start streaming index file. We will collect index file into `result_buf` and use that to write the index into disk. For now we assume that the data sent will be in little endian byte ordering.

Corresponding PR in Lantern Extras [PR](https://github.com/lanterndata/lantern_extras/pull/142)

### TODO
- [x] add tests

### DIFF
--- a/scripts/integration_tests.py
+++ b/scripts/integration_tests.py
@@ -663,12 +663,14 @@ def external_index(request):
             tries = 1
             time.sleep(5)
 
-@pytest.mark.parametrize("distance_metric", ["l2sq", "cos"])
-@pytest.mark.parametrize("quant_bits", [32, 16, 8])
+@pytest.mark.parametrize("distance_metric", ["l2sq", "cos", "hamming"])
+@pytest.mark.parametrize("quant_bits", [32, 16, 8, 1])
 @pytest.mark.external_index
 def test_external_index(external_index, primary, source_table, quant_bits, distance_metric):
-    table_name = f"{source_table}_{quant_bits}_external_index"
+    table_name = f"{source_table}_{quant_bits}_{distance_metric}_external_index"
     index_name = f"idx_hnsw_{table_name}_{distance_metric}"
+
+    data_type = 'INT[]' if distance_metric == 'hamming' else 'REAL[]'
 
     if not primary.execute(
         "testdb",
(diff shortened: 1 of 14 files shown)


src/hnsw/build.c
```diff
@@ -510,7 +675,20 @@ static void BuildIndex(Relation heap, Relation index, IndexInfo *indexInfo, ldb_
                  estimated_row_count,
                  "index size exceeded maintenance_work_mem during index construction, consider increasing "
                  "maintenance_work_mem");
-        usearch_reserve(buildstate->usearch_index, estimated_row_count, &error);
+
+        if(buildstate->external) {
+            assert(is_little_endian());
```
Ngalstyan4, line 680: Make this "if" and elog error on fail, so the check happens in release builds

src/hnsw/build.c
```diff
@@ -540,23 +763,27 @@ static void BuildIndex(Relation heap, Relation index, IndexInfo *indexInfo, ldb_
...
     assert(error == NULL);
     buildstate->usearch_index = NULL;
 
-    fstat(index_file_fd, &index_file_stat);
-    result_buf = mmap(NULL, index_file_stat.st_size, PROT_READ, MAP_PRIVATE, index_file_fd, 0);
-    assert(result_buf != MAP_FAILED);
+    if(!buildstate->external) {
+        fstat(index_file_fd, &index_file_stat);
+        result_buf = mmap(NULL, index_file_stat.st_size, PROT_READ, MAP_PRIVATE, index_file_fd, 0);
+        assert(result_buf != MAP_FAILED);
```
Ngalstyan4, line 785: Make this elog error 
Q: is it safe to elog error from this context ?

src/hnsw/build.c
```diff
@@ -522,14 +700,59 @@ static void BuildIndex(Relation heap, Relation index, IndexInfo *indexInfo, ldb_
...
+        result_buf = palloc0(index_size);
+
+        assert(result_buf != NULL);
+
+        // start reading index into buffer
+        while(total_received < index_size) {
+            bytes_read = read(buildstate->external_client_fd, result_buf + total_received, BUFFER_SIZE);
+
+            // Check for CTRL-C interrupts
+            CHECK_FOR_INTERRUPTS();
```
Ngalstyan4, line 743: try or test this

scripts/integration_tests.py
```diff
@@ -645,6 +648,95 @@ def test_vector_search_with_filter(primary, source_table):
...
+            "testdb",
+            f"CREATE TABLE IF NOT EXISTS {table_name} AS SELECT * FROM {source_table}",
+        )
+        primary.execute("testdb", f"ALTER TABLE {table_name} ADD PRIMARY KEY (id)")
+
+    ops = "dist_l2sq_ops" if distance_metric == "l2sq" else "dist_cos_ops"
+    # note: when M is too low, the streaming search below might return fewer results since the bottom layer of hnsw is less connected
+    primary.execute(
+        "testdb",
+        f"CREATE INDEX {index_name} ON {table_name} USING lantern_hnsw (v {ops}) WITH (dim=128, M=10, quant_bits = {quant_bits}, external = true)",
```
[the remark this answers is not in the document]
var77: yes this gets executed for 8, 16, 32 quant bits and cos, l2sq distances

PR #325, opened 2024-07-24
Update extension version checking logic

var77:
Generate static array from CMake for compatible version list based on update file names. Use that array to check if current catalog version of extension is compatible with current binary version


src/hnsw/utils.c
```diff
@@ -135,77 +138,91 @@ bool VersionsMatch()
...
-        }
+    if(isnull) {
+        elog(ERROR, "SQL version query returned null");
+    }
+
+    // Grab the result and check that it matches the version in the generated header
+    version_text = DatumGetTextP(val);
+    catalog_version = text_to_cstring(version_text);
+
+    if(strlen(catalog_version) == 0) {
```
Ngalstyan4, line 194: so, if catalog version has zero length, we assume it is always compatible with any binary version?
is it what we want? I would think we would consider this error in all cases.

src/hnsw/utils.c
```diff
@@ -135,77 +138,91 @@ bool VersionsMatch()
...
-        if(comparison == 0) {
-            versions_match = true;
-        }
+    if(isnull) {
+        elog(ERROR, "SQL version query returned null");
+    }
+
+    // Grab the result and check that it matches the version in the generated header
+    version_text = DatumGetTextP(val);
+    catalog_version = text_to_cstring(version_text);
```
Ngalstyan4, line 192: You are storing `version_text` in a global variable but I am pretty sure it gets `free()`ed after the current context is gone, while `catalog_version` remains available as a freed pointer in the global process scope.

Is `catalog_version` actually used anywhere/
if so, it should probably be a NAME-sized static array into which you can `strncpy` the value of `version_text`

src/hnsw/utils.c
```diff
@@ -22,8 +22,11 @@
 #include "usearch_storage.hpp"
 #include "version.h"
 
-bool versions_match = false;
-bool version_checked = false;
+bool  version_checked = false;
+bool  versions_match = true;
```
Ngalstyan4, line 26: why default versions_match to true here?

PR #327, opened 2024-08-02
External indexing socket with SSL support

var77:


scripts/integration_tests.py
```diff
@@ -652,13 +653,18 @@ def test_vector_search_with_filter(primary, source_table):
 @pytest.fixture
 def external_index(request):
     cli_path = os.getenv("LANTERN_CLI_PATH")
+    use_ssl = os.getenv("EXTERNAL_INDEX_SECURE") == "1"
```
Ngalstyan4, line 656: Can this variable be called `USE_SSL` or whatever we decide is the uniform name of the variable?

src/hnsw/external_index_socket.c
```diff
@@ -108,6 +119,24 @@ static int connect_with_timeout(int sockfd, const struct sockaddr *addr, socklen
...
+int sendall(external_index_socket_t *socket_con, const char *buf, uint32 len, int flags)
+{
+    int total = 0;
+    int bytesleft = len;
+    int n;
+
+    while(total < len) {
+        n = socket_con->send(socket_con, buf + total, bytesleft, flags);
+        if(n == -1 || n == 0) {
+            return n;
```
Ngalstyan4, line 131: why is this silently returning the number and not closing the socket and erroring out?

it seems we would want to error out from all the callsites anyway. If so, the API is consuming since the caller might not realize they have to check for these values and deal with closing/erroring out since otherwise the function is quite high level (it handles partial sends and retires)

src/hnsw/external_index_socket.c
```diff
@@ -217,43 +278,45 @@ int create_external_index_session(const char                   *host,
...
+    check_external_index_request_error(socket_con, bytes_written);
 
     if(params->pq) {
         external_index_send_codebook(
-            client_fd, buildstate->pq_codebook, params->dimensions, params->num_centroids, params->num_subvectors);
+            socket_con, buildstate->pq_codebook, params->dimensions, params->num_centroids, params->num_subvectors);
     }
 
-    uint32 buf_size = read(client_fd, &init_response, EXTERNAL_INDEX_INIT_BUFFER_SIZE);
+    uint32 buf_size = socket_con->read(socket_con, (char *)&init_response, EXTERNAL_INDEX_INIT_BUFFER_SIZE);
```
Ngalstyan4, line 290: buf_size type must must int32 since it might report -1 an later error handling function below compares it against -1

PR #328, opened 2024-08-16
Support sparsevec in weighted vector search function

dqii:


test/sql/weighted_search.sql
```diff
@@ -121,6 +120,30 @@ SELECT count(*)
   w3=> 0.52, col3=>'v_real'::text, vec3=>:'v444'::vector
 );
 
+-- Check if the type 'sparsevec' exists and store the result in a variable
+SELECT EXISTS (
```
Ngalstyan4, line 124: could we follow the current pattern and not do this in the test itself?

Maybe, see how we filter out pgvector or lantern extras tests when relevant extensions are not installed and do something similar here

test/sql/weighted_search.sql
```diff
@@ -121,6 +120,30 @@ SELECT count(*)
...
+    \ir utils/small_world_sparsevec.sql
+    SELECT '{1:0.4,2:0.3,3:0.2}/3' AS s3 \gset
+    SELECT '[-0.5,-0.1,-0.3]' AS v3 \gset
+    SELECT
+      id,
+      0.9 * (s <-> :'s3'::sparsevec) + 0.1 * (v <-> :'v3'::vector) as dist
+    FROM lantern.weighted_vector_search(CAST(NULL as "small_world"), exact => false, ef => 5,
+      w1=> 0.9, col1=>'s'::text, vec1=>:'s3'::sparsevec,
+      w2=> 0.1, col2=>'v'::text, vec2=>:'v3'::vector
+    );
```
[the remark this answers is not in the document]
var77: Should we do something like this?
```sql
round(cast(0.9 * (s <-> :'s3'::sparsevec) + 0.1 * (v <-> :'v3'::vector) as numeric), 2) as dist
```

On arm processors the high precision floats seems to slightly differ from x86 ones which makes tests to fail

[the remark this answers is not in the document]
dqii: Fixed^

docker/Dockerfile.dev
```diff
@@ -31,7 +31,7 @@ RUN gem install pg -- --with-pg-include=/usr/local/pgsql/include/ --with-pg-lib=
 # hack to make sure postgres user has write access to externally mounted volumes
 RUN mkdir /lantern_shared && chown postgres:postgres /lantern_shared
 
-RUN cd /root/postgresql-15.5/contrib && make install -j
+RUN cd /root/postgresql-15.5/contrib && make install
```
Ngalstyan4, line 34: Why change this?

docker/Dockerfile.dev
```diff
@@ -55,7 +55,7 @@ COPY . .
 RUN sudo rm -rf build \
     && mkdir build \
     && cd build \
-    && cmake -DCMAKE_BUILD_TYPE=Debug .. \
+    && cmake .. \
```
Ngalstyan4, line 58: why change this?

The dockerfile is meant for development and I often use GDB from inside it, so having a build with symbols is often good.

Though, I usually attach a folder from host and rebuild the DB for debugging so not a big deal but would still like to know why the change.

dqii: The build did not succeed for me without it

docker/Dockerfile.dev
dqii: The build did not succeed for me without it

dqii: 440.5 /usr/bin/install -c -m 755  postgres_fdw.so '/usr/local/pgsql/lib/postgres_fdw.so'
440.5 /usr/bin/install -c -m 644 ./postgres_fdw.control '/usr/local/pgsql/share/extension/'
440.6 /usr/bin/install -c -m 644 ./postgres_fdw--1.0.sql ./postgres_fdw--1.0--1.1.sql  '/usr/local/pgsql/share/extension/'
440.6 make[1]: Leaving directory '/root/postgresql-15.5/contrib/postgres_fdw'
------
Dockerfile.dev:34
--------------------
  32 |     RUN mkdir /lantern_shared && chown postgres:postgres /lantern_shared
  33 |     
  34 | >>> RUN cd /root/postgresql-15.5/contrib && make install -j
  35 |     
  36 |     # allow non-root users to install in the container to make it easier to run update-tests
--------------------
ERROR: failed to solve: process "/bin/sh -c cd /root/postgresql-15.5/contrib && make install -j" did not complete successfully: exit code: 2

View build details: docker-desktop://dashboard/build/desktop-linux/desktop-linux/ojyc3i59rx9nve3a7qeqpdid7
diqi@Dis-Laptop lantern % 

The error I get with -j

PR #332, opened 2024-09-05
change palloc to malloc for external indexing

var77:
Palloc has limit of 1GB ([source code reference](https://github.com/postgres/postgres/blob/1b373aed20e61e4a3033e1e396e4ba7c2a96bc20/src/include/utils/memutils.h#L40-L42)) , so for large indexes it was failing to allocate a necessary buffer to store the index.


var77:
Covered by https://github.com/lanterndata/lantern/pull/333

PR #333, opened 2024-09-08
Refactor `StoreExternalIndexNodes` function and receive index file from socket in chunks

var77:
Refactor StoreExternalIndexNodes function and receive index file from socket in chunks

- simplify `StoreExternalIndexNodes` function to write the provided buffer into index pages until fully flushed.
- try to write as much nodes as possible before requesting new chunk from external indexing server to optimize data streaming
- do not throw errors or process interrupts from `StoreExternalIndex` or external index socket functions, instead set status and error msg to `buildstat->status`, then check the status and throw the error or process interrupts after resource cleanup.
- receive and process index file in 10MB chunks

### DIFF
--- a/src/hnsw/external_index_socket.c
+++ b/src/hnsw/external_index_socket.c
@@ -15,7 +15,7 @@ static bool is_little_endian()
     return *((char *)&i) == 1;
 }
 
-static void set_read_timeout(int32 client_fd, uint32 seconds)
+static void set_read_timeout(int32 client_fd, uint32 seconds, BuildIndexStatus *status)
 {
     struct timeval timeout;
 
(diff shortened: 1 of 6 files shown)


src/hnsw/external_index_socket.c
```diff
@@ -132,22 +149,19 @@ static int connect_with_timeout(int sockfd, const struct sockaddr *addr, socklen
...
-        elog(ERROR, "external index socket read failed");
+    switch(check_external_index_response(socket_con, buffer, size)) {
+        case EXTERNAL_INDEX_NO_ERR:
+            return;
+        case EXTERNAL_INDEX_READ_FAILED:
+            socket_con->close(socket_con);
+            elog(ERROR, "external index socket read failed");
+            break;
+        case EXTERNAL_INDEX_INDEXING_ERROR:
+            buffer[ size ] = '\0';
```
Ngalstyan4, line 160: what is the guarantee that this does not overflow the `buffer`?

PR #341, opened 2024-10-14
add openai completion API to SQL and daemon

var77:
- Add SQL function to call OpenAI completion API
- Refactor embedding jobs, remove CSV exporter, make db_exporter more generic


lantern_extras/default_15658317961189010738_0_46722.profraw
dqii, line 1: Was this intended to be included?

var77: it was by mistake thanks for noticing!

PR #347, opened 2024-10-29
Narek/update deps

Ngalstyan4:
This updates (almost) all rust dependencies of the project

Dependencies left out of date:
1. gcp_auth - An authentication API changed, did not upgrade since I did not know the new API and the right way to test
2. ndarray - upgrading causes some cryptic error messages

Notes on pgrx upgrade:

- The added `[[bin]]` section and the `pgrx_embed.rs` file were required by `pgrx 0.12.*`. Those would be autogenerated if we created a pgrx project with the new version.
- I dropped `pg11` but have not added `pg17` yet. Let's do that separately. I just needed pgrx 0.12.* for other changes.

### DIFF
--- a/lantern_cli/src/http_server/pq.rs
+++ b/lantern_cli/src/http_server/pq.rs
@@ -32,7 +32,7 @@ pub struct CreatePQInput {
     post,
     path = "/collections/{name}/pq",
     request_body  (
-        content = CreateIndexInput,
+        content = CreatePQInput,
         example = json!(r#"{ "column": "vector", "clusters": 2, "splits": 1 }"#),
     ),
     responses(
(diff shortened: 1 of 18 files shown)


lantern_cli/src/http_server/pq.rs
```diff
@@ -1,3 +1,4 @@
+use crate::http_server::index::CreateIndexInput;
```
var77, line 1: I think this was initially wrong, in the route instead of `CreateIndexInput` it should have been `CreatePQInput` which is declared in the file, but for the previous versions maybe macro didn't expanded correctly and it didn't fail the build

Ngalstyan4: fixed

PR #348, opened 2024-10-29
pg17 support

var77:


lantern_hnsw/scripts/test_updates.py
```diff
@@ -49,7 +49,8 @@ def __repr__(self):
         return self.version
 
 INCOMPATIBLE_VERSIONS = {
-    '16': [Version('0.0.4')]
+    '16': [Version('0.0.4')],
+    '17': [Version('0.3.0'), Version('0.3.1'), Version('0.3.2'), Version('0.3.3'), Version('0.3.4'), Version('0.4.0'), Version('0.4.1')],
```
Ngalstyan4, line 53: If we no longer have servers at an older `0.3.*` version that we need to upgrade from, you can also move old update scripts to [old_updates](https://github.com/lanterndata/lantern/tree/main/lantern_hnsw/sql/old_updates)

lantern_hnsw/test/expected/hnsw_vector.out
```diff
@@ -199,48 +184,43 @@ CREATE INDEX l2_idx ON small_world_arr USING lantern_hnsw(v) WITH (dim=3, m=2);
...
+SELECT id FROM small_world_arr ORDER BY v <=> ARRAY[0,0,0];
  id 
 ----
   1
   2
   3
-(3 rows)
 
 DROP INDEX cos_idx;
 CREATE INDEX ham_idx ON small_world_arr USING lantern_hnsw(v) WITH (m=3);
```
Ngalstyan4, line 213: was this supposed to be testing hamming distance? it seems the index creation uses the default l2sq operator in this case as well - there is a typo in the old test.

Ngalstyan4 (approved):

lantern_hnsw/test/expected/hnsw_vector.out
var77: fixed to create an index with correct opclass

lantern_hnsw/scripts/test_updates.py
var77: there are still servers with 0.3.x versions, will move the update files to old_updates once all servers are upgraded

PR #355, opened 2024-11-04
external index server in bgworker

var77:
- Add external indexing server as background worker process in Lantern Extras
- Deprecate index from file functionality in lantern and throw a message to use the new version
- Remove old external indexing functionality from lantern/lantern_extras/lantern_cli
- Remove tests associated with old external indexing functionality
- Set `lantern.external_index_secure=false` by default, so it will correctly connect to background worker.


lantern_extras/src/lib.rs
```diff
@@ -97,6 +122,73 @@ pub extern "C" fn lantern_daemon_worker() {
...
+            latch
+        };
+
+        if (wakeup_flags & pg_sys::WL_POSTMASTER_DEATH as i32) != 0 {
+            break;
+        }
+        /* ========================================================== */
+
+        if BackgroundWorker::sighup_received() && !ENABLE_INDEXING_SERVER.get() {
+            std::process::exit(1);
```
Ngalstyan4, line 167: Why are we exiting with an error code when the external indexing is not enabled?
afaik this causes a postmaster restart, no? is that expected?

lantern_hnsw/src/hnsw.c
```diff
@@ -425,21 +425,9 @@ Datum       lantern_internal_failure_point_enable(PG_FUNCTION_ARGS)
...
 
-// todo:: remove in 0.3.1
-PGDLLEXPORT PG_FUNCTION_INFO_V1(lantern_internal_continue_blockmap_group_initialization);
-Datum       lantern_internal_continue_blockmap_group_initialization(PG_FUNCTION_ARGS)
-{
-    LDB_UNUSED(fcinfo);
-    PG_RETURN_NULL();
-}
-
+// todo:: remove in 0.4.3
```
Ngalstyan4, line 428: should be 0.5.1 or 0.6.0

lantern_hnsw/src/hnsw/build.c
```diff
@@ -486,6 +486,12 @@ static void BuildIndex(Relation heap, Relation index, IndexInfo *indexInfo, ldb_
 
     InitBuildState(buildstate, heap, index, indexInfo);
 
+    if(buildstate->index_file_path) {
+        elog(ERROR,
+             "Importing index from file is deprecated.\n"
```
Ngalstyan4, line 491: deprecated makes it sound like the functionality still exists, but we have removed the functionality already.
Rephrase to "Importing index from file is no longer supported"

Ngalstyan4 (approved):
small questions, but overall looks good.

lantern_extras/src/lib.rs
var77: It is because if the bgworker exits with 0 status it won't be restarted automatically and will be unregistered from postmaster. By exiting the process we make sure that any threads created by this process will be killed and as it exits with non-zero code, it will be restarted after the configured `bgw_restart_time`. After boot it won't start the indexing server (as GUC will be set to false) and will wait for SIGHUP in wait_latch loop, to start in case the `lantern_extras.enable_indexing_server` will be set to true again.

Ref to docs: https://www.postgresql.org/docs/current/bgworker.html#:~:text=If%20bgw_restart_time%20for,itself%20has%20terminated.

